PR #1, opened 2020-12-19
Initial refactor of napari 780

sofroniewn:
After further discussion with @jni we decided that the napari animation functionality best belongs as a plugin rather than in the main repo, and I have begun this repo and done some refactoring of https://github.com/napari/napari/pull/780 to use some of the new `asdict` functionality in our dataclasses.

Thanks for all your work on this last year @guiwitz and I'm sorry it took so long for us to come back to this. Having this in a new repo now though will mean we can do much more rapid iteration before we release.

One change I've made is to make an `Animate` class that has no Qt or keybindings, and is meant for purely programatic animation making, and to make an `AnimateWidget` class that can be added as a dock widget, and then as a plugin once https://github.com/napari/napari/pull/1849 merges.

Here's a look at the current animation widget

Screen Shot 2020-12-19 at 3 46 17 PM

I have dreams of making a `QListWidget` so you can see the keyframes you've already captured! Maybe @tlambert03 has ideas here too.

You can get a sense of what is possible now by looking in the examples folder. Another change is that between each pair of keyframes you can have a different number of interpolation steps, and you can add an "easing" function that can be used to adjust the transition - see [easing functions web](https://easings.net/) and [easing functions in python](https://github.com/semitable/easing-functions)

Right now it will only capture changes in `camera` or `dims`, but we can expand this with time.

Also, I really want to push on type checking/ coercion for our models so that my interpolation stuff can leverage it better - it's not so great right now for int/ floats - see https://github.com/napari/napari/issues/2009 on the main repo for some initial discussions of how we might do that. We probably need to make some improvements in napari there before we can release this widely.


LICENSE
```diff
@@ -1,7 +1,15 @@
+BSD 3-Clause License
+
+From napari-animation:
 
 Copyright (c) 2020, Nicholas Sofroniew
 All rights reserved.
 
+From naparimovie:
```
sofroniewn, line 8: cc @guiwitz I added this copy right notice to reflect the input from naparimovie

napari_animation/animation.py
```diff
@@ -1,334 +1,171 @@
...
-            )
+        new_state = {
+            'camera': self.viewer.camera.asdict(),
+            'dims': self.viewer.dims.asdict(),
+        }
 
-        # update view
-        self.viewer.window.qt_viewer.view.camera.view_changed()
+        # Log transform zoom for linear interpolation
+        new_state['camera']['zoom'] = np.log10(new_state['camera']['zoom'])
```
sofroniewn, line 86: I don't love that i have to do this, I wonder if we want to store the log transformed zoom on the camera instead, or if there is a cleaner way to do this ... can figure out later though

examples/animate3D.py
```diff
@@ -0,0 +1,33 @@
...
+    animation.capture_keyframe()
+    viewer.camera.zoom = 2.4
+    animation.capture_keyframe()
+    viewer.camera.angles = (-7.0, 15.7, 62.4)
+    animation.capture_keyframe(steps=60)
+    viewer.camera.angles = (2.0, -24.4, -36.7)
+    animation.capture_keyframe(steps=60)
+    viewer.reset_view()
+    viewer.camera.angles = (0.0, 0.0, 90.0)
+    animation.capture_keyframe()
```
sofroniewn, line 32: @GenevieveBuckley I imagine we'll use syntax like this where we explicitly call `animation.capture_keyframe` if we want to autogenerate animations as part of our docs!

napari_animation/animation.py
```diff
@@ -1,334 +1,171 @@
...
-        implot : matplotlib Ax object
-            reference to matplotlib image used for movie returned by imshow
-        anim : matplotlib FuncAnimation object
-            reference to animation object
+        steps : int
+            Number of interpolation steps between last keyframe and captured one.
+        ease : callable, optional
+            If provided this method should make from `[0, 1]` to `[0, 1]` and will
+            be used as an easing function for the transition between the last state
+            and captured one.
```
sofroniewn, line 42: Here's a nice [gist of easing functions in python](https://gist.github.com/zeffii/c1e14dd6620ad855d81ec2e89a859719). We could expose a subset of these as an ENUM in a dropdown menu for GUI users

PR #2, opened 2020-12-27
Add easing functions and Easing Enum for GUI

alisterburt:
This PR adds 

- the easing functions from https://gist.github.com/zeffii/c1e14dd6620ad855d81ec2e89a859719
- an Enum for selecting them from the GUI

The only thing changed in the functions were occasional addition or removal of parentheses, enforcing a consistent coding style and adding simple docstrings from info in the gist

In the enum, I chose to only include the functions which ease both in and out, but left the rest in the module in case there's a reason for wanting single sided easing that I'm missing

I decided against using the napari StringEnum object because napari wasn't a requirement of this project

What I think I should do before merge:
- add tests to check that functions only generate values between 0 and 1 for inputs between 0 and 1
- hook up a GUI easing function selector (would be great to get a little guidance on this)

Look forward to hearing your thoughts!

### DIFF
--- a/napari_animation/easing.py
+++ b/napari_animation/easing.py
@@ -1,0 +1,285 @@
+"""
+The easing functions here were adapted from
+https://gist.github.com/zeffii/c1e14dd6620ad855d81ec2e89a859719
+"""
+
+from math import sqrt, pow, sin, cos, pi
+from enum import Enum
+
+tau = pi * 2
+
+
+class Easing(Enum):
+    """Easing: easing function to use for a transition.
+
+    Selects a preset easing function
+            * linear: linear interpolation between start and endpoint.
+            * quadratic: quadratic easing in and out.
+                Modeled after the piecewise quadratic
+                y = (1/2)((2x)^2)             ; [0, 0.5)
+                y = -(1/2)((2x-1)*(2x-3) - 1) ; [0.5, 1]
+            * cubic: cubic easing in and out.
+                Modeled after the piecewise cubic
+                y = (1/2)((2x)^3)       ; [0, 0.5)
+                y = (1/2)((2x-2)^3 + 2) ; [0.5, 1]
+            * quintic: quintic easing in and out.
+                Modeled after the piecewise quintic
+                y = (1/2)((2x)^5)       ; [0, 0.5)
+                y = (1/2)((2x-2)^5 + 2) ; [0.5, 1]
+            * sine: sinusoidal easing in and out.
+                Modeled after half sine wave
+                y = 0.5 * (1 - cos(x * pi))
+            * circular: circular easing in and out.
+                Modeled after the piecewise circular function
+                y = (1/2)(1 - sqrt(1 - 4x^2))           ; [0, 0.5)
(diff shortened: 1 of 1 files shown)


napari_animation/easing.py
```diff
@@ -0,0 +1,285 @@
...
+https://gist.github.com/zeffii/c1e14dd6620ad855d81ec2e89a859719
+"""
+
+from math import sqrt, pow, sin, cos, pi
+from enum import Enum
+
+tau = pi * 2
+
+
+class Easing(Enum):
```
sofroniewn, line 12:
```suggestion
class Easing(str, Enum):
```

napari_animation/easing.py
```diff
@@ -0,0 +1,285 @@
...
+    LINEAR = 1
+    QUADRATIC = 2
+    CUBIC = 3
+    QUINTIC = 4
+    SINE = 5
+    CIRCULAR = 6
+    EXPONENTIAL = 7
+    ELASTIC = 8
+    BACK = 9
+    BOUNCE = 10
```
sofroniewn, line 59: Can you make these of the form `LINEAR='linear'` as it means people can then use string for Enum along with above change, see https://stackoverflow.com/a/63028809

napari_animation/easing.py
```diff
@@ -0,0 +1,285 @@
+"""
+The easing functions here were adapted from
+https://gist.github.com/zeffii/c1e14dd6620ad855d81ec2e89a859719
+"""
```
sofroniewn, line 4:
```suggestion
original c code:
https://raw.githubusercontent.com/warrenm/AHEasing/master/AHEasing/easing.c
Copyright (c) 2011, Auerhaus Development, LLC
http://sam.zoy.org/wtfpl/COPYING for more details.
"""
```
Just adding this, but all the license is maximally permissive.

Link above takes you too
```
            DO WHAT THE FUCK YOU WANT TO PUBLIC LICENSE
                    Version 2, December 2004

 Copyright (C) 2004 Sam Hocevar <sam@hocevar.net>

 Everyone is permitted to copy and distribute verbatim or modified
 copies of this license document, and changing it is allowed as long
 as the name is changed.

            DO WHAT THE FUCK YOU WANT TO PUBLIC LICENSE
   TERMS AND CONDITIONS FOR COPYING, DISTRIBUTION AND MODIFICATION

  0. You just DO WHAT THE FUCK YOU WANT TO.
```

sofroniewn (approved):
Looks good to me @alisterburt! I've made a couple minor requests, and then I'll merge.

We can then update #1 to use these functions!

napari_animation/easing.py
sofroniewn: fine to keep all caps on the enum part and go lower case on the string part

napari_animation/easing.py
alisterburt: hahahah fantastic - looks great!

napari_animation/easing.py
alisterburt: done and done! thanks!

alisterburt:
okay, thanks for your suggestions, learned a few things!
should be good to go :rocket:

sofroniewn (approved):
LGTM! Great work!!

PR #5, opened 2021-01-09
Hook up easing combobox

sofroniewn:
Hi @alisterburt this PR begins to connect a combobox to the easing functions that you added in #2. I'm using the Enum that you provided, but instead of it being a string enum we probably want to store the functions inside, like in this magicgui example https://github.com/napari/magicgui/blob/master/examples/napari_image_arithmetic.py#L11-L22. See some discussion here as well https://github.com/napari/napari/pull/2079#issuecomment-755636552.

If you'd like to take over this PR I can help you through it. It should be fun!

Also if you're interesting in contributing more to this repo I'd love that, especially around the `AnimationWidget` I think there's a lot we can do there to make it better!!

### DIFF
--- a/napari_animation/_qt/animation_widget.py
+++ b/napari_animation/_qt/animation_widget.py
@@ -74,17 +74,24 @@ class AnimationWidget(QWidget):
     def _get_interpolation_steps(self):
         return int(self.frameWidget.stepsSpinBox.value())
 
+    def _get_easing_function(self):
+        easing_name = str(self.frameWidget.easeComboBox.currentText())
+        #TODO insert logic to go from name to easing function and return function
+        # instead of None 
+        print('easing', easing_name)
+        return None
+
     def _set_current_frame(self):
         return self.frameWidget.frameSpinBox.setValue(self.animation.frame)
 
     def _capture_keyframe_callback(self, event=None):
         """Record current key-frame"""
-        self.animation.capture_keyframe(steps=self._get_interpolation_steps())
+        self.animation.capture_keyframe(steps=self._get_interpolation_steps(), ease=self._get_easing_function())
         self._set_current_frame()
 
     def _replace_keyframe_callback(self, event=None):
         """Replace current key-frame with new view"""
-        self.animation.capture_keyframe(steps=self._get_interpolation_steps(), insert=False)
+        self.animation.capture_keyframe(steps=self._get_interpolation_steps(), ease=self._get_easing_function(), insert=False)
         self._set_current_frame()
 
     def _delete_keyframe_callback(self, event=None):
--- a/napari_animation/_qt/animation_widget.py
+++ b/napari_animation/_qt/animation_widget.py
@@ -1,6 +1,7 @@
 from qtpy.QtWidgets import QWidget, QLabel, QVBoxLayout, QLineEdit, QPushButton
 
 from ..animation import Animation
+from ..easing import Easing
 from .frame_widget import FrameWidget
 
 
@@ -22,7 +23,6 @@ class AnimationWidget(QWidget):
     def __init__(self, viewer: 'napari.viewer.Viewer', parent=None):
         super().__init__(parent=parent)
 
-        
         self._layout = QVBoxLayout()
         self.setLayout(self._layout)
 
@@ -76,17 +76,16 @@ class AnimationWidget(QWidget):
 
     def _get_easing_function(self):
         easing_name = str(self.frameWidget.easeComboBox.currentText())
-        #TODO insert logic to go from name to easing function and return function
-        # instead of None 
-        print('easing', easing_name)
-        return None
+        easing_func = Easing[easing_name.upper()].value
+        return easing_func
 
     def _set_current_frame(self):
         return self.frameWidget.frameSpinBox.setValue(self.animation.frame)
 
     def _capture_keyframe_callback(self, event=None):
         """Record current key-frame"""
-        self.animation.capture_keyframe(steps=self._get_interpolation_steps(), ease=self._get_easing_function())
+        self.animation.capture_keyframe(steps=self._get_interpolation_steps(),
+                                        ease=self._get_easing_function())
         self._set_current_frame()
 
     def _replace_keyframe_callback(self, event=None):
@@ -121,6 +120,6 @@ class AnimationWidget(QWidget):
         print('Saving animation to', path)
         self.animation.animate(path)
 
-    def close():
+    def close(self):
         self._release_callbacks()
         super().close()
--- a/napari_animation/_qt/frame_widget.py
+++ b/napari_animation/_qt/frame_widget.py
@@ -1,4 +1,7 @@
-from qtpy.QtWidgets import QWidget, QFormLayout, QSpinBox
+from qtpy.QtCore import Qt
+from qtpy.QtWidgets import QComboBox, QWidget, QFormLayout, QSpinBox
+
+from ..easing import Easing
 
 
 class FrameWidget(QWidget):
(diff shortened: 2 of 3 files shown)


alisterburt:
Hey Nick, thanks for reaching out! 

I had a quick play with providing the functions in the enum and I'm running into weird issues, do you have any ideas?

```python
In[2]: runfile('/home/aburt/programming/napari-animation/napari_animation/easing/easing.py', wdir='/home/aburt/programming/napari-animation/napari_animation')
In[3]: class Easing(str, Enum):
  ...:     LINEAR = 'linear'
  ...:     QUADRATIC = 'quadratic'
  ...:     CUBIC = 'cubic'
  ...:     QUINTIC = 'quintic'
  ...:     SINE = 'sine'
  ...:     CIRCULAR = 'circular'
  ...:     EXPONENTIAL = 'exponential'
  ...:     ELASTIC = 'elastic'
  ...:     BACK = 'back'
  ...:     BOUNCE = 'bounce'
  ...: 
  ...: 
In[4]: Easing
Out[4]: <enum 'Easing'>
In[5]: for f in Easing: print(f)
  ...: 
Easing.LINEAR
Easing.QUADRATIC
Easing.CUBIC
Easing.QUINTIC
Easing.SINE
Easing.CIRCULAR
Easing.EXPONENTIAL
Easing.ELASTIC
Easing.BACK
Easing.BOUNCE
In[6]: class Easingf(Enum):
  ...:     LINEAR = linear_interpolation
  ...:     QUADRATIC = quadratic_ease_in_out
  ...:     CUBIC = cubic_ease_in_out
  ...:     QUINTIC = quintic_ease_in_out
  ...:     SINE = sine_ease_in_out
  ...:     CIRCULAR = circular_ease_in_out
  ...:     EXPONENTIAL = exponential_ease_in_out
  ...:     ELASTIC = elastic_ease_in_out
  ...:     BACK = back_ease_in_out
  ...:     BOUNCE = bounce_ease_in_out
  ...: 
In[7]: Easingf
Out[7]: <enum 'Easingf'>
In[8]: for f in Easingf: print(f)
  ...: 

```

but when I do the magicgui example the same way

```python
class Operation(Enum):
    add = numpy.add
    subtract = numpy.subtract
    multiply = numpy.multiply
    divide = numpy.divide
    
for f in Operation: print(f)
Operation.add
Operation.subtract
Operation.multiply
Operation.divide
```

Using partial in the `Easingf` version does yield a valid enum

```python
class Easingpf(Enum):
    LINEAR = partial(linear_interpolation)
    QUADRATIC = partial(quadratic_ease_in_out)
    CUBIC = partial(cubic_ease_in_out)
    QUINTIC = partial(quintic_ease_in_out)
    SINE = partial(sine_ease_in_out)
    CIRCULAR = partial(circular_ease_in_out)
    EXPONENTIAL = partial(exponential_ease_in_out)
    ELASTIC = partial(elastic_ease_in_out)
    BACK = partial(back_ease_in_out)
    BOUNCE = partial(bounce_ease_in_out)
    
for f in Easingpf: print(f)
Easingpf.LINEAR
Easingpf.QUADRATIC
Easingpf.CUBIC
Easingpf.QUINTIC
Easingpf.SINE
Easingpf.CIRCULAR
Easingpf.EXPONENTIAL
Easingpf.ELASTIC
Easingpf.BACK
Easingpf.BOUNCE

```

I checked whether having the functions in a separate package made a difference, same behaviour. Is there something special about the numpy functions which makes them work here whilst the easing functions don't? Why does partial fix this?

Possible solution:
Store a dict which maps the keys to the functions, create the enum dynamically from that dict in the `AnimationWidget`? After having a quick read I assume you want the enum as an enum for persistence later (read [here](http://xion.io/post/code/python-enums-are-ok.html))

I'm definitely keen to contribute, I just tend to worry that if I hack things together it's not going to be quite right for the *bigger picture*, I guess there's only one way to learn? :laughing:

sofroniewn:
> I checked whether having the functions in a separate package made a difference, same behaviour. Is there something special about the numpy functions which makes them work here whilst the easing functions don't? Why does partial fix this?

Hmm, this it's strange to me that the numpy functions work! You can read more about why `partial` is needed here https://stackoverflow.com/a/40339397, but it's not clear to me why the numpy functions work. It would be much nicer if we were in that situation. Maybe @tlambert03 knows? I'll look into this now a little more too. I'd prefer to use the Enum approach and not the dict approach if possible.

> I'm definitely keen to contribute, I just tend to worry that if I hack things together it's not going to be quite right for the bigger picture, I guess there's only one way to learn? 😆

That's great! And `napari-animation` will be a good place to learn as we can iterate a little faster here as it's not a dependency of the main repo

tlambert03:
> Hmm, this it's strange to me that the numpy functions work! 

yeah, I never figured that out either!  This recently came up in napari too: https://github.com/napari/napari/pull/2079#issuecomment-755878115

I can dig a little deeper to figure out what's going on with numpy ufuncs if you don't figure it out in the meantime.  I'm curious.

tlambert03:
one thing that probably matters is that `np.add` (and the other `ufuncs`) are _not_ of type `function`... they are instances of the class `numpy.ufunc`.  So ... something silly like this works for enums:

```python
In [1]: from enum import Enum

In [2]: class FunnyFunc:
   ...:     def __init__(self, func):
   ...:         self.func = func
   ...:
   ...:     def __call__(self, *args, **kwargs):
   ...:         return self.func(*args, **kwargs)
   ...:
   ...:     @property
   ...:     def __signature__(self):
   ...:         import inspect
   ...:         return inspect.signature(self.func)
   ...:

In [3]: def func(x):
   ...:     print(x)
   ...:

In [4]: funny_func = FunnyFunc(func)

In [5]: class Test(Enum):
   ...:     func = func
   ...:     funny_func = funny_func
   ...:

In [6]: list(Test)
Out[6]: [<Test.funny_func: <__main__.FunnyFunc object at 0x7fd2800801c0>>]

In [7]: vars(Test)
Out[7]:
mappingproxy({'_generate_next_value_': <function enum.Enum._generate_next_value_(name, start, count, last_values)>,
              '__module__': '__main__',
              'func': <function __main__.func(x)>,
              '__doc__': 'An enumeration.',
              '_member_names_': ['funny_func'],
              '_member_map_': {'funny_func': <Test.funny_func: <__main__.FunnyFunc object at 0x7fd2800801c0>>},
              '_member_type_': object,
              '_value2member_map_': {<__main__.FunnyFunc at 0x7fd2800801c0>: <Test.funny_func: <__main__.FunnyFunc object at 0x7fd2800801c0>>},
              'funny_func': <Test.funny_func: <__main__.FunnyFunc object at 0x7fd2800801c0>>,
              '__new__': <function enum.Enum.__new__(cls, value)>})
```

alisterburt:
huh, weird! I think this is because when the value is a function it's actually no different than a method of the `Enum` 

```python
In [13]: class FuncEnum(Enum):
    ...:     a = f
    ...:     def b():
    ...:         pass
    ...: 

In [14]: FuncEnum.a
Out[14]: <function __main__.f(x)>

In [15]: FuncEnum.b
Out[15]: <function __main__.FuncEnum.b()>
```

I'll try to find a nice clean way around it so we don't have to have `partial(f)` everywhere

tlambert03:
> I think this is because when the value is a function it's actually no different than a method of the `Enum`

exactly... as described in that SO post that @sofroniewn posted: https://stackoverflow.com/a/40339397

sofroniewn:
> one thing that probably matters is that np.add (and the other ufuncs) are not of type function... they are instances of the class numpy.ufunc. So ... something silly like this works for enums:

ah nice, that probably explains this.

> I'll try to find a nice clean way around it so we don't have to have partial(f) everywhere

I guess given the complexity of the other solutions, having `partial` might actually be the cleanest way forward, and I think will meet our needs. I think I prefer using `partial` to defining our own custom class.

@alisterburt I'd recommend just using partial and then wiring everything up appropriately. `[e.name.lower() for e in Easing]` gives you the lowercase names and `Easing[name.upper()].value` will give you a callable from a lowercase name. (I have a slight preference for upper case names in the Enum and lowercase names in the combobox, but it's pretty arbitrary and maybe that's just me).

Feel free to make a PR to this PR, and let me know if you run into additional problems or come up with a better solution!

alisterburt:
sounds great, I'm fine with doing it that way! will drop in a PR soon (and sorry for missing the SO link!)

tlambert03:
> I think I prefer using partial to defining our own custom class

yeah def (partial is basically doing something very similar to `FunnyFunc` there anyway).

sofroniewn:
This is now working! Thanks all!!

PR #7, opened 2021-01-10
connect up easing functions to animation widget

alisterburt:
All linked up with the Easing enum for function selection!

This is really fun, I had forgotten dims weren't included in state so got to poke around in the state generation/interpolation machinery too when debugging, very cool.

Here's a little video with the exponential easing

https://user-images.githubusercontent.com/7307488/104137269-831dd380-5393-11eb-82db-f39afe90036a.mp4

Edit: looks like dims should be supported, wasn't working in my hands when making the animation interactively (where camera worked fine). Will have a look

### DIFF
--- a/napari_animation/_qt/animation_widget.py
+++ b/napari_animation/_qt/animation_widget.py
@@ -1,6 +1,7 @@
 from qtpy.QtWidgets import QWidget, QLabel, QVBoxLayout, QLineEdit, QPushButton
 
 from ..animation import Animation
+from ..easing import Easing
 from .frame_widget import FrameWidget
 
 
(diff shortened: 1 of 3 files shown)


sofroniewn (approved):
This looks great! Thanks!!

> Edit: looks like dims should be supported, wasn't working in my hands when making the animation interactively (where camera worked fine). Will have a look

This might be because of problems with the type based interpolation right now. We look at whether something is an int/ float and decide how to interpolate based on that, but right now we have no gaurentee on what types something is. After https://github.com/napari/napari/pull/2042 merges we'll have much better control over types

PR #14, opened 2021-01-12
[WIP] key frame list widget

alisterburt:
still very much work in progress but thought it might be worth checking in!

I currently have it set up so that 
- `KeyFramesListWidget` generates a label and an id for the viewer state whenever a keyframe is captured.
- The label is displayed in the `QListWidget`
- The label is stored in a dict mapping id to label
- the `KeyFramesListWidget._update_from_animation()` method can repopulate the list with the correct labels based on the current state of `Animation.key_frames` (i.e. upon insertion/deletion/reordering but this doesn't happen automatically yet)

Next steps are to set up mechanisms for modifying the list from the GUI. This took me waaaay too long but I'm already learning a bunch!

Main questions are around the id -> label mapping dict, does this seem like a decent approach to you?

### DIFF
--- a/napari_animation/__init__.py
+++ b/napari_animation/__init__.py
@@ -1,3 +1,3 @@
 from .animation import Animation
 from ._qt import AnimationWidget
-from ._hookimpls import napari_experimental_provide_dock_widget
+# from ._hookimpls import napari_experimental_provide_dock_widget
--- a/napari_animation/__init__.py
+++ b/napari_animation/__init__.py
@@ -1,3 +1,3 @@
 from .animation import Animation
 from ._qt import AnimationWidget
-# from ._hookimpls import napari_experimental_provide_dock_widget
+from ._hookimpls import napari_experimental_provide_dock_widget
--- a/napari_animation/_qt/keyframeslist_widget.py
+++ b/napari_animation/_qt/keyframeslist_widget.py
@@ -4,7 +4,7 @@ from qtpy.QtCore import QSize
 
 from napari.utils.events import EventedList
 
-# NOT IMPLEMENTED YET
+
 class KeyFramesListWidget(QListWidget):
     """List of key frames for an animation
 
@@ -18,6 +18,7 @@ class KeyFramesListWidget(QListWidget):
     animation : napari_animation.Animation
         napari-animation animation in sync with the GUI.
     """
+
     def __init__(self, animation, parent=None):
         super().__init__(parent=parent)
 
@@ -208,7 +209,8 @@ class KeyFramesListWidget(QListWidget):
     def backend_key_frame_labels(self):
         """labels for key frames in the order present in the backend
         """
-        return [self._key_frame_id_to_label_map[id(key_frame)] for key_frame in self.animation.key_frames]
+        return [self._key_frame_id_to_label_map[id(key_frame)] for key_frame in
+                self.animation.key_frames]
 
     @property
     def labels_to_key_frames(self):
--- a/napari_animation/_qt/keyframeslist_widget.py
+++ b/napari_animation/_qt/keyframeslist_widget.py
@@ -49,6 +49,7 @@ class KeyFramesListWidget(QListWidget):
     def _selection_callback(self, event):
         self._update_frame_number()
         self.animation.set_to_current_keyframe()
+        self.parentWidget()._update_frame_widget_from_animation()
 
     def _add(self):
         """Generate QListWidgetItem for current keyframe, store its unique id and add it to self
@@ -61,7 +62,6 @@ class KeyFramesListWidget(QListWidget):
         self._map_key_frame_to_id(key_frame_idx)
         self._map_label_to_qlistwidgetitem(key_frame_idx)
         self._frame_counter += 1
-        self._update_visibility()
 
     def insertItem(self, row, item):
         """overrides QListWidget.insertItem to also update current index and frame number
@@ -76,13 +76,6 @@ class KeyFramesListWidget(QListWidget):
         self.takeItem(event.index)
         self._remove_key_frame_id(event.value)
         self._update_frame_number()
-        self._update_visibility()
-
-    def _update_visibility(self):
-        if len(self.animation.key_frames) >= 1:
-            self.show()
-        else:
-            self.hide()
 
     def _remove_key_frame_id(self, key_frame):
         """Remove key-frame id from self._key_frame_id_to_label_map
--- a/napari_animation/_qt/keyframeslist_widget.py
+++ b/napari_animation/_qt/keyframeslist_widget.py
@@ -52,7 +52,6 @@ class KeyFramesListWidget(QListWidget):
     def _add(self):
         """Generate QListWidgetItem for current keyframe, store its unique id and add it to self
         """
-        print(self.animation.frame)
         key_frame_idx = self.animation.frame
         item = self._generate_list_item(key_frame_idx)
 
@@ -63,6 +62,7 @@ class KeyFramesListWidget(QListWidget):
         self._map_key_frame_to_id(key_frame_idx)
         self._map_label_to_qlistwidgetitem(key_frame_idx)
         self._frame_counter += 1
+        self._update_visibility()
 
     def _remove(self, event):
         """Remove QListWidgetItem at event.index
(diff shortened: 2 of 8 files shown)


sofroniewn:
> the KeyFramesListWidget._update_from_animation() method can repopulate the list with the correct labels based on the current state of Animation.key_frames (i.e. upon insertion/deletion/reordering but this doesn't happen automatically yet)

I wonder here if we might need to use something like the event system from napari and our evented list. You can get it via `from napari.utils.events import EventedList`, see https://github.com/napari/napari/blob/master/napari/components/layerlist.py. The nice thing about it is that it will emit "events" that you can listen to on the Qt side. You can read more about events here https://napari.org/docs/dev/events/event_loop.html

> Main questions are around the id -> label mapping dict, does this seem like a decent approach to you?

That seems reasonable. You might not need it in the long run if we do with the event system above, but I don't have any problems with it.

> Next steps are to set up mechanisms for modifying the list from the GUI. This took me waaaay too long but I'm already learning a bunch!

This sounds good. That's great that you're learning here! I think focusing on getting the GUI side nice first and then we can hook up the other direction using the event system probably makes the most sense! Nice work

alisterburt:
perfect, thanks for the feedback!
I'll play around with the implementation I had in mind and try the evented
list, I suspect the evented list will make my life easier :)

alisterburt:
@sofroniewn tidied this up a bit, really fun! 
- drag and drop in the gui now works and appropriately reorders things in the `Animation.key_frames` list
- I made the `key_frames` list an `EventedList` 

For now, my mechanism for updating the front end based on back end state makes no distinctions about what type of list mutation is occuring (except insertion), it just regenerates the list on a change

I guess thumbnails is next!

alisterburt:
@sofroniewn and thumbnails are working! :tada: 

Here's what they look like in the viewer

https://user-images.githubusercontent.com/7307488/104777716-0fedd600-5774-11eb-916b-3ff2eba42340.mp4



and here's what the rendered animation looks like

https://user-images.githubusercontent.com/7307488/104777695-03697d80-5774-11eb-8961-bcec421187ff.mp4

tlambert03:
holy shit!

sofroniewn:
This is incredible!! I'll try and do some review tonight, great progress. Curious your thoughts on what should be next/ what's still left on this PR?

alisterburt:
Thanks both of you!! 

the main issue I see is that if the GUI updates when the backend changes, you lose the thumbnails. 

I think I'd like to move thumbnail generation onto the backend to sort this out, putting it in the key_frames dict like the easing function.

The other thing was moving it to the right place. I can't remember why I put it at the bottom right now, something about the animation not being instantiated yet..

Once those things are sorted I guess it's ready for merge but I'd like to keep iterating, get the parameter editing working

alisterburt:
I'll ping you for review when it's ready to save you some time!

sofroniewn:
> I think I'd like to move thumbnail generation onto the backend to sort this out, putting it in the key_frame dict like the easing function.

Yes this sounds wise. Good idea!

sofroniewn:
this continues to come along really nicely. I just played some more with it, and have a couple comments/ questions.

curious what you think we should do with the frame combobox from the top section? should we drop it and now just use the "selection" in the list to determine where the next keyframe gets added? it feels a little redundant now that we have the list. if we make the next frame always appear after the currently selected frame.

we should add a `remove frame` button next to the `capture frame` button that would delete the selected keyframe.

can you move the list widget above the textbox for the filename, keeping the save animation at the very bottom.

I think we should make the steps/ ease displayed in the GUI reflect and edit the values of the currently selected frame. on initialization / with an empty list or with none selected they should probably be disabled, then as soon as there is one selected they show the values of the selected one. (This could come in a followup PR).

I noticed that if I rearranged the list in the GUI and then click capture frame nothing happens.

alisterburt:
Thanks for taking a look! I think you're right yeah, I'll remove it and get the selection position working.

I've moved the list widget up. I noticed the same bug when capturing after rearranging, there's also another one where if you rearrange again after rearranging then capturing a keyframe you get a `KeyError` - going to have a look at those now.

Making the parameters directly editable like that sounds like a great solution - I was imagining something more complicated with proper 'selection' for editing on double-click but your way seems more intuitive :+1:

alisterburt:
okay, this is now a working version and I guess ready for review @sofroniewn ! :rocket: 

key points:
- all per key-frame parameters are now editable from the frame widget
- selecting a key-frame moves you to that key frame
- key-frame list is editable/reorderable
- key-frame list has a delete frame button

There are definitely a few bits where it feels like I'm having to work around rather than with Qt, especially when trying to keep various widgets in sync with bits of the backend which don't have events - if you see glaring faults I'm definitely keen to have them pointed out

There's something slightly jarring about the way the KeyFramesLidgetWidget (QListWidget) initialises with a width larger than the current AnimationWidget width - I tried playing with size policies and things but couldn't get that to properly initialise to the current width of the animation widget - any advice there is appreciated

napari_animation/_qt/keyframeslist_widget.py
```diff
@@ -1,10 +1,214 @@
...
+
+    def _update_backend(self):
+        """push current GUI state to self.animation
+        """
+        new_key_frames = [self._label_to_key_frame(label) for label in
+                          self.frontend_key_frame_labels]
+
+        # recreating the EventedList here and reconnecting events simplifies handling events
+        # which fire when we modify the list in place
+        self.animation.key_frames = EventedList(new_key_frames)
```
sofroniewn, line 101: Hmm yeah, this seems undesirable - it's not quite clear to me why we have to replace the whole list and we can't just reorder the elements and get events from this

napari_animation/animation.py
```diff
@@ -26,12 +27,12 @@ class Animation:
     def __init__(self, viewer):
         self.viewer = viewer
 
-        self.key_frames = []
+        self.key_frames = EventedList()
         self.frame = -1
```
sofroniewn, line 31: In long run we could make `frame` be evented too, so you could connect to an event whenever this changed? Would that help? Probably something we can save for a new PR though

napari_animation/__init__.py
```diff
@@ -1,3 +1,3 @@
 from .animation import Animation
 from ._qt import AnimationWidget
-from ._hookimpls import napari_experimental_provide_dock_widget
+# from ._hookimpls import napari_experimental_provide_dock_widget
```
sofroniewn, line 3: Did this need to be commented? I will uncomment.

sofroniewn (approved):
@alisterburt this looks great! I just gave the functionality a try and it works really well!!

> There are definitely a few bits where it feels like I'm having to work around rather than with Qt, especially when trying to keep various widgets in sync with bits of the backend which don't have events - if you see glaring faults I'm definitely keen to have them pointed out

In the code the only thing around the events that I see that really seems off is having to create a new EventedList. I think we'll want to figure out how to avoid that, and we might want to add a new event that gets emitted when the `frame` changes too, but we can do those in new PRs.

> There's something slightly jarring about the way the KeyFramesLidgetWidget (QListWidget) initialises with a width larger than the current AnimationWidget width - I tried playing with size policies and things but couldn't get that to properly initialise to the current width of the animation widget - any advice there is appreciated

I will give this a look right now. I might also switch to using enabled/ disabled on the widgets rather than show/ hide as it could make that less jarring.

All in all this is great and I think we should get in now. We should probably then add some tests to the repo, setup the CI etc, before doing more to make sure that subsequent changes preserve our functionality

napari_animation/_qt/keyframeslist_widget.py
alisterburt: totally agree - it felt wrong, but I'll make sorting it a priority moving forward! :+1:

alisterburt:
thanks for the review @sofroniewn ! Agree on all points you raised. I'll find some time this week for addressing your comments and moving forward on the testing/CI/deployment side :rocket:

sofroniewn:
@alisterburt I've pushed some styling changes to the widget

Before:
Screen Shot 2021-01-27 at 2 52 01 PM

After:
Screen Shot 2021-01-27 at 2 50 42 PM

If this looks ok to you I will then merge, we can continue to iterate on styles and layouts in follow up PRs. Nice work here!

napari_animation/_qt/keyframeslist_widget.py
sofroniewn: See notes below, I'm happy to merge this as is, it's a big improvement, and then you can work on this in a followup PR

alisterburt:
I love it!! Just had a play around with it locally and it's infinitely better, thanks! lets go! :rocket:

sofroniewn:
> I love it!! Just had a play around with it locally and it's infinitely better, thanks! lets go! 🚀

Great, will merge now. I've also created #18 for us to track some more of the refactors we want to do

PR #19, opened 2021-01-28
add a save dialog

### LINKED ISSUE #15: Make save animation button launch a dialog
We should drop the save string and make the save animation button open up a proper saving dialog box

alisterburt:
basically a straight up copy of the screenshot dialog in napari, seems to work like a charm! Closes #15

### DIFF
--- a/napari_animation/_qt/dialogs/save_dialog.py
+++ b/napari_animation/_qt/dialogs/save_dialog.py
@@ -1,7 +1,7 @@
 from pathlib import Path
 from typing import Any, Callable
 
-from PyQt5.QtWidgets import QFileDialog, QMessageBox
+from qtpy.QtWidgets import QFileDialog, QMessageBox
 
 
 class SaveAnimationDialog(QFileDialog):
--- a/napari_animation/_qt/dialogs/save_dialog.py
+++ b/napari_animation/_qt/dialogs/save_dialog.py
@@ -1,12 +1,12 @@
-import os
-from typing import Any, Callable
 from pathlib import Path
+from typing import Any, Callable
+
 from PyQt5.QtWidgets import QFileDialog, QMessageBox
 
 
 class SaveAnimationDialog(QFileDialog):
     """
-    Dialog to choose save location of animation
+    Dialog to choose the save location of animation
 
     Parameters
     ----------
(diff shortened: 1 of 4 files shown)


napari_animation/_qt/dialogs/save_dialog.py
```diff
@@ -0,0 +1,50 @@
+from pathlib import Path
+from typing import Any, Callable
+
+from PyQt5.QtWidgets import QFileDialog, QMessageBox
```
sofroniewn, line 4: Can this be a qtpy import instead of PyQt5

sofroniewn (approved):
Looks good to me if you change the import to qtpy (we should be using that everywhere in the repo).

@tlambert03, I wonder if you'd ever want to put something in magicgui to assist with open/ save dialogs like this? I didn't see anything in there quite like it right now, but might be something to consider, though it might also be some scope creep too! I can raise an issue on magicgui if you think it warrants further discussion there.

alisterburt:
Perfect, will change the import when I get back to a computer!

napari_animation/_qt/dialogs/save_dialog.py
tlambert03: > @tlambert03, I wonder if you'd ever want to put something in magicgui to assist with open/ save dialogs like this? I didn't see anything in there quite like it right now, but might be something to consider, though it might also be some scope creep too! I can raise an issue on magicgui if you think it warrants further discussion there.

The functionality is there in the [`FileEdit` widget](https://napari.org/magicgui/_autosummary/magicgui.widgets.FileEdit.html#magicgui.widgets.FileEdit), but _just_ opening a dialog should be exposed a little better, and since it's not really a widget, currently requires this nonsense:

```python
from magicgui.application import use_app

show_dialog = use_app().get_obj("show_file_dialog")
file = show_dialog(mode="w")  # returns a string path where they want to save to
```
where "mode" is one of the [`FileDialogModes`](https://napari.org/magicgui/api/types.html#magicgui.types.FileDialogMode)

alisterburt:
okay all sorted there, should be ready to go!

sofroniewn:
Thanks!!

PR #20, opened 2021-01-29
Highlight colour

### LINKED ISSUE #17: Make highlight color same as napari highlight color
It would be nice if the qlistwidget used the same color as the napari layer list for its highlight color (bright blue for the dark theme, instead of gray).

See below:
Screen Shot 2021-01-27 at 2 50 42 PM

alisterburt:
image

this PR closes #17 

I also removed the background to match how the napari layer list looks, if you don't prefer that I'm happy to switch it back!

### DIFF
--- a/napari_animation/_qt/animation_widget.py
+++ b/napari_animation/_qt/animation_widget.py
@@ -27,6 +27,7 @@ class AnimationWidget(QWidget):
 
         # Create animation
         self.animation = Animation(viewer)
+        self.viewer = viewer
 
         self._layout = QVBoxLayout()
         self.setLayout(self._layout)
@@ -45,6 +46,9 @@ class AnimationWidget(QWidget):
         self.saveButton = QPushButton('Save Animation', parent=self)
         self.saveButton.clicked.connect(self._save_callback)
         self._layout.addWidget(self.saveButton)
+
+        self.viewer.events.theme.connect(self._update_theme)
+        self._update_theme()
 
         # establish key bindings
         self._add_keybind_callbacks()
@@ -144,6 +148,22 @@ class AnimationWidget(QWidget):
         path = self.pathText.text()
         print('Saving animation to', path)
         self.animation.animate(path)
+
+    def _update_theme(self, event=None):
+        """Update from the napari GUI theme"""
+        from napari.utils.theme import get_theme, template
+
+        # get theme and raw stylesheet from napari viewer
+        theme = get_theme(self.viewer.theme)
+        raw_stylesheet = self.viewer.window.qt_viewer.raw_stylesheet
+
+        # template and apply the primary stylesheet
+        templated_stylesheet = template(raw_stylesheet, **theme)
+        self.setStyleSheet(templated_stylesheet)
+
+        # update styling of KeyFramesListWidget
+        self.keyframesListWidget._update_theme(theme)
+
 
     def close(self):
         self._release_callbacks()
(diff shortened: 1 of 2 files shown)


napari_animation/_qt/keyframeslist_widget.py
```diff
@@ -185,6 +191,14 @@ def _key_frame_to_label(self, key_frame):
         key_frame_id = id(key_frame)
         return self._key_frame_id_to_label_map[key_frame_id]
 
+    @property
+    def _item_background_color(self):
+        return 'QListView::item:selected{background-color: rgb(0,122,204);}'
```
sofroniewn, line 196: Can you get this color from the napari theme - you might need to do something like @tlambert03 did here https://github.com/napari/napari-console/blob/3313a710ad671532727c6d2f4f8d255afdc6bb05/napari_console/qt_console.py#L151-L155.

and connect an update them method here https://github.com/napari/napari-console/blob/3313a710ad671532727c6d2f4f8d255afdc6bb05/napari_console/qt_console.py#L81.

Not how the import is done inside the `_update_theme` method to cheat a little on our dependencies.

I don't all love this and think we might want to find a better way long term to allow plugins to update stylings based on themes, but this is where we are at now :-)

sofroniewn (commented):
This is looking very nice @alisterburt - I left one comment about not hardcoding the value but using the value from our theme.

Now that I think about it more actually @tlambert03 is there a way we could do this with the inherited stylesheet from napari, maybe with some custom qss? Would be nicer that way.

Also from an overall styling perspective, I like that you got rid of the overall background color for the keyframe list, but can you now add in a color for the unselected items too that is the same as the color in the napari layer list. Basically the more the keyframe list looks like the napari layer list the better I think as people will be visually used to it

alisterburt:
Thanks for the review, appreciate the detailed comments/pointing to how this was solved elsewhere! 

I'll put the same mechanism in place for getting theme info and try to match the LayerList styling. 

Longer term, is the way to improve this separating out the theming of napari into its own plugin (on which napari itself and plugins would depend for theming?)

alisterburt:
image
image

updated things to better match the styling of the napari `LayerList` although there are a few inconsistencies which I've given up on after 30 mins of qss fiddling - I'll open a separate issue for those now because the highlight colour seems to be working okay :rocket:

napari_animation/_qt/animation_widget.py
```diff
@@ -145,6 +150,21 @@ def _save_callback(self, event=None):
...
+
+        # template and apply the primary stylesheet
+        templated_stylesheet = template(raw_stylesheet, **theme)
+        self.setStyleSheet(templated_stylesheet)
+
+        # update styling of KeyFramesListWidget
+        self.keyframesListWidget._update_theme(theme)
+
     def close(self):
         self._release_callbacks()
```
sofroniewn, line 169: Can you add here inside the close method `self.viewer.events.theme.disconnect(self._update_theme)` too

sofroniewn (approved):
Styles look great now, I've added one request to disconnect the viewer connection in the close method, otherwise this looks good to go. no more need to fiddle with the stylings at this point. I think we'll want to make it easier on the plugin developers to do this sort of stuff in the future and this is a great starting point now!!

alisterburt:
done & done, cheers!

sofroniewn:
Awesome! Will merge now!!

alisterburt (commented):
uh oh, messed up fixing the merge conflicts :-1: will sort out and PR again

sofroniewn:
oops, thank's i'll get the CI workflow working soon, won't take me too long

sofroniewn:
actually @alisterburt what isn't working - it looks good for me

alisterburt:
@sofroniewn locally none of the highlight work was showing up, maybe just in a git mess... will try to dig myself out :laughing: 

it's [this commit](https://github.com/sofroniewn/napari-animation/pull/20/commits/e0c8c19ffaf1fc706614b5311f676d4381bb89e8) which seems to have removed it though

Issue #18, opened 2021-01-27
Event refactors


sofroniewn:
Follow up from #14, stop recreating `EventedList` and make the `frame` property evented (note here we might want to wait until after https://github.com/napari/napari/pull/2127 is available as it might make this much cleaner) - cc @alisterburt

alisterburt:
Had a quick go at eventing the frame counter this evening and didn't get too far. How would you go about doing this? Is making the frame attribute into the value of a QSpinBox the completely wrong way to go?

sofroniewn:
This will be much easier after the next napari release that contains https://github.com/napari/napari/pull/2127, but I would then turn [Animation](https://github.com/sofroniewn/napari-animation/blob/b64f943fdc0b76a61a2cb2a569ed5c1ce27dff81/napari_animation/animation.py#L12) into an `EventedModel` (I'd probably make `viewer` a private attribute too), and then you'll get an `animation.events.frame` eventemitter for free and you can connect that into whatever you like. Certainly it should connect into the "selected" aspect of the QLayerList, so that when you change the frame you update which keyframe is selected and visa-versa. I'm not quite sure I see why you want the QSpinBox too, but maybe I'm missing something

alisterburt:
Okay that sounds great, the EventedModel (and pydantic in general) look super useful. Thanks!

The QSpinBox was just the only way I could think of to get a number which emitted events on changing value that I could connect to - probably showing my naivety 😝 the EventedModel will solve the problem perfectly

Issue #16, opened 2021-01-27
Drop skimage dependency


sofroniewn:
It would probably be nice to drop the scikit-image dependency, instead using things like `scipy.ndimage` as napari itself doesn't depend on skimage

alisterburt:
will look at this when refactoring

sofroniewn:
Closed by #31

PR #25, opened 2021-02-01
simplify the KeyFramesListWidget, everything now works on events

alisterburt:
this PR addresses #18 (only for the `KeyFramesListWidget`, not animation.frame) and simplifies the `KeyFramesListWidget` quite a bit.

Everything in the `KeyFramesListWidget` is now based on events and there's no more of that ugly recreation of the `EventedList` when updating things.

I didn't get around to fixing #16 with this but it's on the list!


sofroniewn (approved):
Looks good to me!

Issue #8, opened 2021-01-11
Configure tests!


sofroniewn:
We should configure tests in the plugin to now use github actions like the main napari repo, cc @ziyangczi. I'm not quite sure what I got from the cookiecutter right now

sofroniewn:
Closed by #26

PR #26, opened 2021-02-01
Get our CI workflow working

sofroniewn:
Get's our CI workflow working - cc @alisterburt


sofroniewn:
We have a working CI now @alisterburt :-) We can then start expanding our tests!

sofroniewn:
Closes #8

PR #30, opened 2021-02-02
Highlight colour reintegration

### LINKED ISSUE #27: Make thumbnails/ list item bigger
I just noticed now that the list widget became very small and i now have a hard time seeing the thumbnails, possibly something that happened in #20 that I missed.

Is that the same for you @alisterburt? Can you make it bigger again, ideally same size as the layer list in napari

Screen Shot 2021-02-01 at 3 49 56 PM

alisterburt:
This PR will close #27, reintegrating the styling work done in #20 which was lost during a broken merge commit


sofroniewn:
This looks great too, I might try a stylings followup to get the padding/ margins a little better around the icon - see now

Screen Shot 2021-02-02 at 6 52 37 PM

Issue #28, opened 2021-02-02
Release 0.0.1 on PyPI


sofroniewn:
Before releasing I'd like us to

- [ ] Close #29 by adding more comprehensive tests
- [x] Close #16 and drop the skimage dependency
- [x] Close #11 which will require napari 0.4.5 (including https://github.com/napari/napari/pull/2042)
- [ ] Close #9 and improve the readme

As we get close to the release I'll also move this over to live at `napari/napari-animation` instead of my personal account, but I want it to get a little bit further along before then.

cc @alisterburt

alisterburt:
Looks great, will work through the list this week!

alisterburt:
Haven't added specific tests on what type of data is coming for each property from the viewer but playing with the latest napari everything seems to be working smoothly, closing #11 (might need to up the minimum napari version?)

There's been progress towards #29 in #34 - can definitely be further improved

#9 is at least partially addressed by #32, I might add a sentence about adding the dock widget from the plugin menu too though

sofroniewn:
Ok nice, sounds like we're making good progress here, would be great to here from @guiwitz other things we want to see before our first release!

guiwitz:
It wasn't really on the list, but I made an attempt #43 at fixing the interactive saving which was non-functional in several aspects. I think this is essential before release. Let me know what you think of the current solution. It's pretty much just fixes to make it work as it was supposed to (I think).

sofroniewn:
Any further thoughts on this @alisterburt @guiwitz? I think we're in a good place to release. Maybe we do some final checks and then try to get `0.0.1` out next week? We can still keep working on it to make it better too :-)

alisterburt:
Just updated #40 - I think it would be worth getting that in for 0.0.1 so people can show/hide layers during an animation, otherwise I think we're good to go! 🚀

guiwitz:
I just pushed a PR #45 that I had planned to submit earlier that adds a scroller to smoothly move through the animation. If you have time to review, it would be nice to have in the first release, but we can also work on that later.

sofroniewn:
Nice! Let's try and get both #40 and #45 in for our `0.0.1`. I'll get to some reviews soon

sofroniewn:
Ok I think we're all done and ready to go for `0.0.1`. I can try and make a release candidate that we can test and then we can release!!

sofroniewn:
[...]

PR #31, opened 2021-02-02
Drop scikit-image dependency

alisterburt:
Dropped the scikit-image dependency!

- used the `napari.utils.io.imsave` implementation which itself uses imageio
- used 'scipy.ndimage.zoom` in place of `skimage.transform` functions
- added an imageio dependency

this goes some of the way to the 0.01 release (#28 )


sofroniewn:
Looks really nice! Great work. I've added `scipy` as a dependency too!

PR #33, opened 2021-02-03
improve margins on thumbnail

sofroniewn:
This PR improves the margins and pads with background the thumbnails

Before:

Screen Shot 2021-02-02 at 6 52 37 PM

After:

Screen Shot 2021-02-02 at 7 51 10 PM

cc @alisterburt


sofroniewn:
I couldn't quite figure out how to get rid of that last gray stripe on the left, can look into that later

sofroniewn:
let me know if you prefer this @alisterburt?

Issue #29, opened 2021-02-02
Add more comprehensive tests


sofroniewn:
Our CI is now working but we should add more comprehensive tests including of the non-qt stuff, and some minimal qt stuff

PR #34, opened 2021-02-07
add tests for animation.py

alisterburt:
first PR towards #29 - some simple tests for the main functionality in animation.py
doesn't include anything for state interpolation yet but given that I'm finishing for the evening and that's in a separate module I thought that could go in a separate PR

I'm a bit worried about how slowly the tests run - this is probably because of my fixtures, am I doing something obviously wrong there?


sofroniewn:
> I'm a bit worried about how slowly the tests run - this is probably because of my fixtures, am I doing something obviously wrong there?

Hmm, we might want to revisit this in a couple days after the `0.4.5` napari release which will provide a `make_napari_viewer` fixture that plugins can use - see https://github.com/napari/napari/pull/2131 for some more detail. I think that can be used "headless" too.

Maybe @tlambert03 has some additional ideas here too

tlambert03:
will dig deeper later.  but my first thought here is that you shouldn't actually be directly testing external libraries and functions like ndi.zoom, and `imageio.get_writer().append_data` so much.  You only need to test _your_ code... not other libraries, so [mock those objects](https://docs.python.org/3/library/unittest.mock.html#quick-guide) and use `assert_called_with` to make sure that your code calls the necessary external functions with the appropriate arguments.

If you want to include a sanity check somewhere that the third party libs are still doing what you want them to be doing.  do it only once or twice, (e.g. with a single frame or something).

alisterburt:
I'd never come across mocking objects, very useful, thanks @tlambert03 ! I'll get back to this soon

tlambert03:
here's a start:
```python
from unittest.mock import patch


@patch('napari_animation.animation.imsave')
@pytest.mark.parametrize('ext', ['.mp4', '.mov', ''])
def test_animate(imsave, animation_with_keyframes, ext):
    """Test that Animation.animate() produces files"""
    animation_with_keyframes.animate(f"test{ext}")
    expected = [f"test_{i}.png" for i in range(30)]
    saved_files = ([call[0][0] for call in imsave.call_args_list])
    assert saved_files == expected
```

currently this is failing on the "no extension" case, but I think that's a legit bug.  Also I think you can probably mock another thing or two to speed it up more

alisterburt:
thanks for the example @tlambert03 ! I quickly saw a couple of different ways to do the mocking so having a concrete (and very clean!) way of doing it is nice, much appreciated!

tlambert03:
> thanks for the example @tlambert03 ! I quickly saw a couple of different ways to do the mocking so having a concrete (and very clean!) way of doing it is nice, much appreciated!

yeah, it's "simple" in concept... but often tricky in practice.  Feel free to reach out if you run into issues

sofroniewn (approved):
Just looked through all these. Tests look great @alisterburt thanks! I'm happy to merge now unless you want to change anything. We can always make follow up PRs to add more too!

napari_animation/_tests/fixtures.py
```diff
@@ -0,0 +1,39 @@
+import pytest
+
+from napari_animation import Animation
+
+@pytest.fixture
+def make_napari_viewer():
```
tlambert03, line 6: curious why you're copying this here?  you already have napari in your test dependencies, and it will provide this

napari_animation/_tests/test_animation.py
```diff
@@ -1,26 +1,90 @@
...
-        return viewer
+import numpy as np
+import pytest
 
-    yield actual_factory
+from napari_animation import Animation
 
-    for viewer in viewers:
-        viewer.close() 
+from .fixtures import make_napari_viewer, empty_animation, animation_with_keyframes, viewer_state
```
tlambert03, line 8: fyi: the "standard" way to use fixtures is either to declare them directly in the file in which you need to use them, or put them into a [conftest.py file](https://docs.pytest.org/en/stable/fixture.html#conftest-py-sharing-fixtures-across-multiple-files) for use across multiple files.  (this way _seems_ to work, but you almost never see it for some reason, and pytest doesn't document it?)

napari_animation/_tests/fixtures.py
sofroniewn: ah, maybe this was before 0.4.5 was released, where I think things were a little in flux (or maybe i just didn't know where things were) we should update this now though, agreed

alisterburt: I'll make sure to import it - thanks!

napari_animation/_tests/test_animation.py
alisterburt: this is really good to know - I guess sticking to the way things should be done does have its benefits - will fix up!

alisterburt:
thanks @sofroniewn and @tlambert03 for taking the time to look this over and sorry for not getting back sooner! Will change a couple of bits tomorrow evening based on the review so we can get this merged then add more in a follow up PR or two

alisterburt:
moved fixtures into conftest, the auto discovery when it's done that way is quite nice! I also added some tests for the interpolation utility functions

napari_animation/_tests/fixtures.py
```diff
@@ -1,26 +1,11 @@
 import pytest
+from napari.utils._testsupport import make_napari_viewer  # noqa: F401
```
sofroniewn, line 2: Love it!

sofroniewn (approved):
Ok, this all looks great! We can keep adding more tests as we go, nice to have this in now!!

napari_animation/_tests/fixtures.py
tlambert03: Here too, I'm not sure whether this might cause you issues.  Fixtures, for better or worse, are not designed to be imported explicitly.  If you have napari installed, pytest will find this fixture (import not necessary).  Because of how terribly ugly that is, we tend to add a comment next to where the fixture is being used, explaining where it comes from, but not actually importing it

sofroniewn: looks like this will be fixed in #42, thanks for the tip!

Issue #11, opened 2021-01-11
Make sure type based interpolation works


sofroniewn:
Right now this plugin is best used with this PR https://github.com/napari/napari/pull/2042 from napari which does type enforcement on our models using pydantic. Until that is released in napari, this plugin won't work so well

sofroniewn:
https://github.com/napari/napari/pull/2042 has now merged, which gives us a deprecation warning on the `asdict` method, we will need to switch to `dict`. We should now test that the interpolation works better. My hope is that it should because pydantic is enforcing correct types for all model values!! cc @alisterburt

sofroniewn:
We've now added some good tests for this

PR #38, opened 2021-03-05
fix deprecated asdict() call

alisterburt:
as mentioned in #11 the `Viewer.dims.asdict()` method is now deprecated - updated to `Viewer.dims.dict()` to fix

### DIFF
--- a/napari_animation/animation.py
+++ b/napari_animation/animation.py
@@ -93,8 +93,8 @@ class Animation:
         """
 
         new_state = {
-            'camera': self.viewer.camera.asdict(),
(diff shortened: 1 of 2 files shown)


sofroniewn (approved):
LGTM!

Do we also need to bump the napari version to be '>0.4.5' for this to always work @alisterburt ? I think maybe, in which case can you do that in this PR too. Thanks!!

alisterburt:
updated the napari req, should be good to go, thanks!!

sofroniewn:
Love it, thanks!

PR #40, opened 2021-03-06
add ability to animate layer attributes

alisterburt:
This PR adds some machinery for getting and setting layer state, allowing changes in layer attributes to be animated.

I tried to write this in a general way so it can be easily extended, for now the only attributes included for animation are `Layer.visible`, `Layer.opacity` and `Layer.blending` which seem to work just fine!

https://user-images.githubusercontent.com/7307488/110219132-567ac980-7eb5-11eb-80bf-5f54236166af.mp4

### DIFF
--- a/napari_animation/animation.py
+++ b/napari_animation/animation.py
@@ -118,27 +118,11 @@ class Animation:
         self.viewer.dims.update(state['dims'])
         self._set_layer_state(state['layers'])
 
-    @property
-    def layers(self):
-        return self.viewer.layers
-
-    def _layer_names(self):
-        return [layer.name for layer in self.layers]
-
-    @property
-    def layer_attributes(self):
-        attributes = (
-            'visible',
-            'opacity',
-            'blending',
-        )
-        return attributes
-
     def _get_layer_state(self):
         """Store layer state in a dict of dicts {layer.name: state}
         """
         layer_state = {
-            layer.name: layer._get_base_state() for layer in self.layers
+            layer.name: layer._get_base_state() for layer in self.viewer.layers
         }
         # remove metadata from layer_state dicts
         for state in layer_state.values():
--- a/napari_animation/animation.py
+++ b/napari_animation/animation.py
@@ -96,6 +96,7 @@ class Animation:
         new_state = {
             'camera': self.viewer.camera.dict(),
             'dims': self.viewer.dims.dict(),
+            'layers': self._get_layer_state(),
         }
 
         # Log transform zoom for linear interpolation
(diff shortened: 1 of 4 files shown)


sofroniewn:
Nice, great to be starting this off! In the long run the layers will be pydantic models and have `dict()` and `json()` methods that can help with capturing state or serialization. Right now there is a `layer._get_state()` method that we could use - see https://github.com/napari/napari/blob/42d48e3d63aa850ec18a3a6167e7441275c94303/napari/layers/image/image.py#L768-L790 and https://github.com/napari/napari/blob/42d48e3d63aa850ec18a3a6167e7441275c94303/napari/layers/base/base.py#L645-L664

We could use these. They are used in our plugin API right now - the writer plugins get access to the "state" dict. They don't ever have the layer `data`, which is generally the large thing, though for some layers like points they can have properties that are specified at the point level, like size or color and so could contain many items, and if we start capturing this every keyframe it could all get very large. They were intentionally left private as it's not quite clear how we wanted them to look and they will be replaced by the pydantic methods.

So I guess all of that was a long way of saying I think I prefer your approach naming the ones we know we want to capture here, but you might find it useful to look through those links.

We might also want to think about saving "diffs" one day instead of saving full snapshots, but that makes it a little harder to do things like rearrange the keyframes or trivially insert or delete, so saving the full ones is still nice.

alisterburt:
thanks for the links/info @sofroniewn - I didn't know about these methods. Once layers are pydantic models, these will be removed fairly quickly I assume? If that's the case then I think explicit is the way to go for now and I'll add in a bunch of extra keys from what's available in the base state/different layer types. 

Once the pydantic layer model is in place I can run some tests and see whether the penalty is too large for keeping track of things like per-point properties. Love the idea of storing diffs, definitely one to keep in mind!

sofroniewn:
> Once layers are pydantic models, these will be removed fairly quickly I assume? 

Yes, but it will take some time before than happens, at least 1-2 months being realistic I think.

> If that's the case then I think explicit is the way to go for now and I'll add in a bunch of extra keys from what's available in the base state/different layer types.

Yup this sounds good, let's just keep an eye out for performance, both the size of the state dict, and the time spend making the movie. One thing we'll get from our evented models is event emission will only happen when the value changes so even if we set the all the attributes new if they are the same as before nothing will happen, which is quite nice!

sofroniewn:
If you end up copying a lot of what's in the layer _get_state methods and you want to use those I think that is fine too, they won't disappear overnight, and we'll be able to update things as needed too

alisterburt:
after seeing a few problems when using `Layer._get_layer_state()` for layer state generation (napari/napari#2480) I dropped this down to using `Layer._get_base_state()` which returns...
- name
- metadata
- scale
- translate
- rotate
- shear
- opacity
- blending
- visible

setting/getting of these properties works appears to work and I think it would be worth getting this in for the 0.0.1 release #28

sofroniewn:
@alisterburt i'm looking through the diff right now on this PR and it seems like there are changes unrelated to this functionality - like in `.pre-commit-config.yaml` and `napari_animation/_qt/dialogs/save_dialog.py`. Can you rebase/ merge master into this PR and then I can review more easily.

> setting/getting of these properties works appears to work and I think it would be worth getting this in for the 0.0.1 release #28

We might want to exclude `metadata` as we don't quite know what will be in it.

Going with the others looks good though. Can you add that list to the readme so that it's clear to users what is and what is not getting recorded. I imagine it might be confusing at first/ until we can support the full list of properties.

Thanks!!

alisterburt:
sorry about that @sofroniewn - added the set of tracked attributes to the readme and removed metadata from the state dict too

all good now and ready for review!

napari_animation/animation.py
```diff
@@ -116,6 +116,40 @@ def _set_viewer_state(self, state):
 
         self.viewer.camera.update(camera_state)
         self.viewer.dims.update(state['dims'])
+        self._set_layer_state(state['layers'])
+
+    @property
```
sofroniewn, line 121: Can we do without this, seems overkill to have `self.layers` when it's pretty easy to get `self.viewer.layers`

napari_animation/animation.py
```diff
@@ -116,6 +116,40 @@ def _set_viewer_state(self, state):
...
+    def layers(self):
+        return self.viewer.layers
+
+    def _layer_names(self):
+        return [layer.name for layer in self.layers]
+
+    @property
+    def layer_attributes(self):
```
sofroniewn, line 129: Can we make this private, I think in the long run as we add things like Image or Points specific attributes we'll probably need to change how we do this, so having private seems better

sofroniewn (approved):
Looks great! Left a couple minor comments. Will be good to merge after that. I do wonder if there is a test we can add for any of this, but that can also come later. Nice work

alisterburt:
Removed some of the cruft around the edges and agree it's much tidier now, thanks for the review @sofroniewn ! , also added a test which checks that the layer properties are captured for animation. Should be good to go!

sofroniewn:
Changes looks good, thanks for that test too - will merge now!!

PR #41, opened 2021-03-09
Fix key-binding to move across frames

guiwitz:
This is a tiny PR which fixes the key-bindings Alt-a and Alt-b. First, because of a bug (missing call to 'viewer' key from the state dictionary) the key-bindings were not functional. Second, once that was fixed I realized that the selection of the key-frame widget row was not updating when moving across key-frames, and this PR fixes that as well.

### DIFF
--- a/napari_animation/_qt/animation_widget.py
+++ b/napari_animation/_qt/animation_widget.py
@@ -140,12 +140,14 @@ class AnimationWidget(QWidget):
 
         new_frame = (self.animation.frame + 1) % len(self.animation.key_frames)
         self.animation.set_to_keyframe(new_frame)
+        self.keyframesListWidget.setCurrentRow(new_frame)
 
     def _key_back_frame(self, event=None):
         """Go backwards in key-frame list"""
 
         new_frame = (self.animation.frame - 1) % len(self.animation.key_frames)
         self.animation.set_to_keyframe(new_frame)
+        self.keyframesListWidget.setCurrentRow(new_frame)
 
     def _save_callback(self, event=None):
         SaveAnimationDialog(self.animation.animate, parent=self).exec_()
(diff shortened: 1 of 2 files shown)


alisterburt (approved):
This is great, thanks for the PR (and your initial napari movie work)!

guiwitz:
Thanks to you and @sofroniewn for turning my initial work into this plugin! It's much nicer to have an actual interface and the handling of interpolation calculation is super sleek now! I'll have a bit of time to work on this and have already implemented things like e.g. a slider to move across key-frames. However I don't want to do PRs without discussing first in what direction you want to go (except to fix bugs like here). Is there a place where to discuss plans for this plugin @sofroniewn ?

alisterburt:
Not trying to speak for Nick but I guess the napari [zulip chat](https://napari.zulipchat.com/) is a good place! Otherwise in the issues here 

Can't wait to hear about your ideas - I was playing with a 'timeline' idea which sounds similar to your slider. Will be fun to discuss!

sofroniewn (approved):
Thanks @guiwitz this looks good! So great to have you contributing to the repo too as all this is based on your original work! There's now a merge conflict, I havn't checked if it's clear how to resolve or not, but if unclear maybe @alisterburt can advise. Thanks!

guiwitz:
It looks like @alisterburt had already fixed part of the problem. I resolved the conflict now. It seem that when I did "git clone", I got the last version of the repository with a commit from @sofroniewn  (this one https://github.com/napari/napari-animation/tree/1fa7b650e5ed3693a2d0a25b4b81f59ed802feae) and not the "real" last one with latest commits from @alisterburt. Is that something that happens?

alisterburt:
@guiwitz it could be that those commits were made a while ago but only recently merged into master? I did have a similar thing recently on another repo though when using the github CLI tool to clone something - in any case thanks for sorting!

PR #42, opened 2021-03-11
conftest.py config + interpolation tests

alisterburt:
There were a couple of commits that got lost from #34 (pushes keep failing when I'm on the VPN, nightmare!) - I've added those here. 

The changes are
- configuring conftest.py for automatic fixture discovery at test time
- dropping the import of `make_napari_viewer` in favour of a comment (thanks @tlambert03 !)


sofroniewn (approved):
Love the interpolation tests. Thanks for cleaning up the fixtures too. This looks good to go!

PR #43, opened 2021-03-18
Fix interactive saving

guiwitz:
This PR fixes several issues with *interactive* saving: 1) it was only possible to save as mp4 because of the pre-selection setting of the dialogue, 2) once that was fixed, when selecting a name without extension, no folder was created to store the PNGs because a wrong file path (with a .png  extension) was passed to the ```animation.animate() ``` function, 3) gifs could not be saved because different imageio writers take different options and a "non-gif" parameter ```quality=5``` was passed by default.

The current behavior is now the following. The user can choose to save with:
- any of the accepted imageio-ffmpeg extensions (.mov, .avi, .mpg, .mpeg, .mp4, .mkv, .wmv) or with the .gif extension. If a wrong extension is used, by default an .mp4 is created.
- a name without extension, in which case, a folder with the given name is created and filled with PNGs

Remaining to be discussed:
- if using a name without extension, one is warned about overwriting. However, only the PNGs with the same name are overwritten. Therefore if one e.g. reduces the number of steps and has a shorter movie, the images of the previous longer movie are still there. Should we remove them actively?
- currently, in the dialogue, one has to select the option to provide either a name for a new folder or a name with an allowed extension. However, one cannot select an *existing* folder. I couldn't find an option to do allow for selection of fil *or* folder [here](https://doc.qt.io/qt-5/qfiledialog.html#FileMode-enum) but maybe someone knows better ?
Screenshot 2021-03-18 at 22 44 55


napari_animation/animation.py
```diff
@@ -225,15 +225,15 @@ def animate(
         if not save_as_folder:
             try:
                 # create imageio writer and add all frames
-                if quality is not None:
+                if path_obj.suffix in ['mov', 'avi', 'mpg', 'mpeg', 'mp4', 'mkv', 'wmv']:
                     writer = imageio.get_writer(
                         path, fps=fps, quality=quality, format=format,
                     )
                 else:
```
alisterburt, line 232: could you add a comment explaining the need for this if/else block, it's not obvious without knowing that gif is the else case why this is happening

alisterburt (approved):
Really nice - I just had a play and everything seems to work the way I would expect it - no errors, defaulted to mp4 without an extension, great job!

I added a tiny comment about the if/else block for handling the `quality` argument to the ffmpeg writer, if you agree could you maybe change that?

I'm realising now that maybe we want to make the save dialog disappear once the file type has been chosen but that can be something we add later.

Re: deletion of extra images in a png folder, I think adding a check and then asking the user what they want is probably the way to go. What do you think?

napari_animation/animation.py
guiwitz: Yes that makes sense. I guess in principle the ```else``` works with anything that's accepted by imageio but not in the ffmpeg-list. But except gif I don't think there's any other possible format. I also updated the parameter definition so that it's clear that several formats other than mp4 are available.

guiwitz:
> I'm realising now that maybe we want to make the save dialog disappear once the file type has been chosen but that can be something we add later.

I'm not sure I understand what you mean here. Doesn't the dialogue disappear once you have chosen a filename?

> Re: deletion of extra images in a png folder, I think adding a check and then asking the user what they want is probably the way to go. What do you think?

Since we already warn the user that he's going to overwrite, I wouldn't actually bother and just delete the content.

sofroniewn:
> Since we already warn the user that he's going to overwrite, I wouldn't actually bother and just delete the content.

Yeah, I think that is fine with me too. As long as the user gets one warning then fine I think

sofroniewn:
> Since we already warn the user that he's going to overwrite, I wouldn't actually bother and just delete the content.

Yeah, I think that is fine with me too. As long as the user gets one warning then fine I think

sofroniewn (approved):
Looks great to me! Thanks. I've left one comment about the folder case and will leave this open till tomorrow if you want to make further adjustments @guiwitz, but then I think it is good to merge :-) Thanks also @alisterburt for the review, so helpful to have had you test it!!

alisterburt:
> Since we already warn the user that he's going to overwrite, I wouldn't actually bother and just delete the content.
Okay cool, let's do that!

guiwitz:
I added a few lines to remove existing PNGs if the selected folder already existed. One last issue: if one selects an already existing folder, one get's the overwriting warning twice. At least on my mac the first time within the OSX UI, and the second time within the napari UI (Qt). I tried to comment out the warning dialogue generation [here](https://github.com/guiwitz/napari-animation/blob/206d2efbe808ef957bd063cbe64e8d994be3add3/napari_animation/_qt/dialogs/save_dialog.py#L38) but then nothing gets saved. I'm not familiar enough with Qt to know what's going on (but it's not too annoying right now).

sofroniewn:
> One last issue: if one selects an already existing folder, one get's the overwriting warning twice. At least on my mac the first time within the OSX UI, and the second time within the napari UI (Qt). I tried to comment out the warning dialogue generation here but then nothing gets saved. I'm not familiar enough with Qt to know what's going on (but it's not too annoying right now).

Maybe @tlambert03 can take a quick look at that, I seem to remember us dealing with something like that in the main repo too. I can try and take a look later today too. We should be able to get this in then! :-)

tlambert03:
yep, you should be able to remove most of the stuff in `accept()`... hang on, working on it

tlambert03:
I would just remove the `save_dialog.py` file altogether and replace your `AnimationWidget._save_callback` with this (adding necessary imports up top):

```python
    def _save_callback(self, event=None):
        filters = (
            "Video files (*.mp4 *.gif *.mov *.avi *.mpg *.mpeg *.mkv *.wmv)"
            ";;Folder of PNGs (*)"  # sep filters with ";;"
        )
        filename, _filter = QFileDialog.getSaveFileName(
            self, "Save animation", str(Path.home()), filters
        )
        if filename:
            self.animate_function(filename)
```

tlambert03:
one other thought: maybe make sure it's going to work (i.e. there are at least 3 key frames, etc...) before allowing them to select a path (which will potentially delete a previously existing thing)

guiwitz:
Thanks for the help @tlambert03 ! I implemented your recommendations. Seems to work better now, the double warning is gone now. I also added an error message if people try to animate less than 2 key frames to avoid needlessly suppressing exiting files.

sofroniewn (approved):
Ok, this looks great! Thanks for the collaboration here, will be nice to get this in and move closer to our first release :-)

PR #45, opened 2021-04-16
Scrollers

guiwitz:
This PR creates a scroller allowing to go through the interpolated animation. To avoid on-the-fly computations of interpolation states, those are calculated upon pressing the "computer interpolation" button. This pre-computing has to be repeated whenever a key-frame is modified, added or removed.

While scrolling through the animation, the selected element in the key-frame list is updated so that one can identify the location of potential issues.


alisterburt:
hey @guiwitz - I love this! Had a play around and it works really nicely, awesome feature!

I'm wondering whether we might want to change the UI around it a little bit - I really like the scroll bar but think it would be much cleaner without the 'compute interpolation' button, what do you think? 

A route to this might be
- storing a bool which keeps track of whether an update is needed
- updating this bool when the keyframeslist changes
- recomputing the interpol states on slider change

There will be a few details which will probably make this tricky and it might feel a little unnatural - do you think it's worth trying to get it to 'just slider' state before 0.0.1? 🚀

guiwitz:
Very good ideas @alisterburt ! I'm not too happy about the compute button either, so I'm totally in favour of improving this. What you propose sounds good. My only fear is what happens when people start having long and complex movies. Will they scroll and nothing will happen because there's computing in the background? Writing this, I realise we have the same problem now: they hit the button but it could take a while until the scroller is usable. Any idea how to handle some kind of a warning about this? We could have some kind of "tqdm"-like widget indicating progress? Have you seen such a thing in other plugins?

guiwitz:
I think this was easier than anticipated and I just implemented what you suggested. I just check whenever a key-frame is added/updated/removed to update the boolean. Whenever the slider moves, it checks the boolean and does the computation if necessary. Much better like this.

alisterburt:
There was [some discussion on the zulip](https://napari.zulipchat.com/#narrow/stream/212875-general/topic/long-running.20computation.20in.20dock.20widget/near/234824916) about this very recently! 

LOVE the updated implementation - glad it worked out! Will look at the code more carefully now

tlambert03:
> We could have some kind of "tqdm"-like widget indicating progress?

there is a tqdm wrapper in magicgui: https://github.com/napari/magicgui/blob/master/magicgui/tqdm.py
but as discussed in that zulip thread, the real challenge is in pushing the computation to another thread

napari_animation/_qt/animation_widget.py
```diff
@@ -40,6 +41,9 @@ def __init__(self, viewer: 'napari.viewer.Viewer', parent=None):
         # Update theme
         self._update_theme()
 
+        # state interpolation updated needed because key-frame changed
+        self.up_to_date_interpolation = False
```
alisterburt, line 45: might be clearer if this slider was an `AnimationSliderWidget` and this flag variable was a property of that object with a name like `is_up_to_date`

napari_animation/_qt/animation_widget.py
```diff
@@ -103,6 +109,10 @@ def _init_save_button(self):
         self.saveButton = QPushButton('Save Animation', parent=self)
         self._layout.addWidget(self.saveButton)
 
+    def _init_scroller_interpol(self):
```
alisterburt, line 112: not immediately obvious from this name what the slider does - see above re: refactor into `AnimationSliderWidget`

napari_animation/_qt/animation_widget.py
```diff
@@ -168,6 +181,23 @@ def _save_callback(self, event=None):
...
+            self._compute_states()
+        new_frame = self.scroller_interpol.value()
+        self.animation._set_viewer_state(self.interpol_states[new_frame])
+        new_key_frame = new_frame // int(self.frameWidget.stepsSpinBox.value())
+        self.keyframesListWidget.setCurrentRow(new_key_frame)
+
+    def _compute_states(self):
+        """Computer interpolation states"""
+        self.interpol_states = []
+        for i, state in enumerate(self.animation._state_generator()):
```
alisterburt, line 196: `i` is unused as far as I can see, might not need the enumerate?

alisterburt (commented):
awesome, thanks for the PR! What do you think about refactoring into a separate subwidget called `AnimationSliderWidget`? Might not be necessary but I think it might make things a little clearer

napari_animation/_qt/animation_widget.py
alisterburt: `requires_update` maybe?

sofroniewn (commented):
This looks great @guiwitz, functionally works as expected for me - though right now I'm unable to access the final state of the movie.

A quick way to check this is to run `python examples/interactive_animation.py` and capture the first slice of the "timeseries" then scroll all the way to the right and capture the last slice, "slice 127".

Using the scroll bar the farthest "right" I can reach takes me to "slice 118" and that second frame never gets highlighted blue. Ideally I should be able to go one more notch to the right, reach "slice 127" and have that second frame get highlighted blue.

Do you think that is possible before merge? Maybe @alisterburt can check this out too.

I also agree with @alisterburt that refactoring into `AnimationSliderWidget` would be nice, also maybe adding a tooltip/ label to explain what the slider is doing in the GUI.

I think this is pretty close though and after we get it in we should be able to make the first release!!

napari_animation/_qt/animation_widget.py
sofroniewn: Agreed, refactoring might be nice/ but we can also do it later after merge. Adding a tooltip / label would probably be helpful too

sofroniewn:
Screen Shot 2021-04-18 at 12 43 33 PM

alisterburt:
@sofroniewn see #47 for why this happens, have opened a separate issue 
That issue should probably be addressed and this should use the fixed state generation machinery leaving us ready to move forward on both - will push a fix when I can

guiwitz:
I refactored the code as suggested. I tired to mirror what you had done for the other widgets in terms of naming conventions. I also included the name changes suggested by @alisterburt. Just be aware that the flag variable ```is_up_to_date``` is now ```requires_update``` which inverts the logic (True is now False). Added a tooltip description to the slider too.

alisterburt:
awesome job @guiwitz - looks great! Let's get this in 😄 🚀

sofroniewn:
Awesome job all, I'll merge now!!

PR #46, opened 2021-04-18
Reduce duplicated widget name

sofroniewn:
Now that napari shows the names of dock widgets this PR drops the extra animation wizard label.

Before:

Screen Shot 2021-04-18 at 1 27 08 PM

After:

Screen Shot 2021-04-18 at 1 24 39 PM

It also changes the example to use `add_plugin_dock_widget`.

@alisterburt can you give this a review/ let me know what you think. Thanks!!

### DIFF
--- a/examples/interactive_animation.py
+++ b/examples/interactive_animation.py
@@ -5,17 +5,13 @@ add_image APIs
 
 from skimage import data
 from scipy import ndimage as ndi
-from napari_animation import AnimationWidget
 import napari
 
 
-with napari.gui_qt():
-    blobs = data.binary_blobs(length=128, volume_fraction=0.1, n_dim=3)
-    viewer = napari.view_image(blobs.astype(float), name='blobs')
-    labeled = ndi.label(blobs)[0]
-    viewer.add_labels(labeled, name='blob ID')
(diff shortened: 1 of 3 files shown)


alisterburt (approved):
This all looks great on the code side - thanks for updating the example to use all the new fancy bits! 

I'm not sure when the plugin widget name label became a part of the dock widget for plugins but I'm not seeing it in my current setup... I'm happy to merge regardless as it doesn't break functionality if the label is missing 
Screenshot 2021-04-18 at 21 32 38

napari: 0.4.7
Platform: macOS-11.2.1-x86_64-i386-64bit
System: MacOS 11.2.1
Python: 3.9.2 | packaged by conda-forge | (default, Feb 21 2021, 05:02:20)  [Clang 11.0.1 ]
Qt: 5.12.9
PyQt5: 5.12.3
NumPy: 1.20.2
SciPy: 1.6.2
Dask: 2021.04.0
VisPy: 0.6.6

sofroniewn:
Ah you might need to be of the master branch of napari then those labels might have just been added. Good to know it doesn't break for you though overall though!! Thanks for the review

Issue #47, opened 2021-04-19
Unreached end state of animation


alisterburt:
as noticed by @sofroniewn in #45 our state interpolation machinery currently doesn't reach the final key-frame of the animation

This happens because we start from the first key frame, and end with a fraction of `n-1 / n` for each pair of keyframes (`n` is the number of steps), this happens on line 146 here
https://github.com/napari/napari-animation/blob/66e9bdf789210ec025a5d61e4dd9fafbeba640c4/napari_animation/animation.py#L138-L151

We can solve this by checking if we're in the last pair of keyframes and either
- adding an extra frame to the animation with the final state
- interpolating from the start to the end state with the same number of frames overall

I'm not sure which of those solutions I prefer, adding an extra frame someone didn't ask for seems like unexpected behaviour (why does my animation have 16 frames when I asked for 15?) but the alternative, not changing the frame number, changes the rate of change slightly and might make things appear faster than they should (could be important for things like time lapse data)

alisterburt:
example code:

```python
from skimage import data
from scipy import ndimage as ndi
from napari_animation import Animation
import napari


blobs = data.binary_blobs(length=128, volume_fraction=0.1, n_dim=3)
viewer = napari.view_image(blobs.astype(float), name='blobs')
labeled = ndi.label(blobs)[0]
viewer.add_labels(labeled, name='blob ID')

animation = Animation(viewer)
viewer.dims.current_step = (0, 0, 0)
animation.capture_keyframe()
viewer.dims.current_step = (127, 0, 0)
animation.capture_keyframe()

for state in animation._state_generator():
    print(state['dims']['current_step'])

```


<details>
<summary>prints</summary>
<br>


```python
(0, 0, 0)
(4, 0, 0)
(8, 0, 0)
(12, 0, 0)
(16, 0, 0)
(21, 0, 0)
(25, 0, 0)
(29, 0, 0)
(33, 0, 0)
(38, 0, 0)
(42, 0, 0)
(46, 0, 0)
(50, 0, 0)
(55, 0, 0)
(59, 0, 0)
(63, 0, 0)
(67, 0, 0)
(71, 0, 0)
(76, 0, 0)
(80, 0, 0)
(84, 0, 0)
(88, 0, 0)
(93, 0, 0)
(97, 0, 0)
(101, 0, 0)
(105, 0, 0)
(110, 0, 0)
(114, 0, 0)
(118, 0, 0)
(122, 0, 0)
```

</details>

[...]

PR #48, opened 2021-04-19
include final key-frame in animation

alisterburt:
fix for #47 - includes a slight refactoring of the state generation method

### DIFF
--- a/napari_animation/_tests/test_animation.py
+++ b/napari_animation/_tests/test_animation.py
@@ -109,3 +109,15 @@ def test_layer_attribute_capture(layer_state, attribute):
     """Test that 'attribute' is captured in the layer state dictionary"""
     for layer_state_dict in layer_state.values():
         assert attribute in layer_state_dict.keys()
+
+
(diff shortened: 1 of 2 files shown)


alisterburt:
also fixes the main issue in #45

sofroniewn (approved):
LGTM! Thanks!!

One minor thing - you need to add a `+1` here https://github.com/napari/napari-animation/blob/66e9bdf789210ec025a5d61e4dd9fafbeba640c4/napari_animation/animation.py#L154 so that the total number of frames is right.

Right now I see
```
Rendering frame  29 of 30
Rendering frame  30 of 30
Rendering frame  31 of 30
```

We might actually want to make the total number of frames a property - something like:

```python
    @property
    def n_frames(self):
        if len(self.key_frames) > 1:
            return np.sum([f["steps"] for f in self.key_frames[1:]]) + 1
        else:
            return 0
```

that could be reused in many places if people want

alisterburt:
@sofroniewn good catch! all fixed based on your suggestions, cheers!

Issue #54, opened 2021-04-25
Setting properties one by one for images triggers many expensive redraws


alisterburt:
I ran into some 'too slow' issues when interactively animating a 3D scene (mentioned in #28) with a 500x500x400 volume

In particular, setting the animation state to a given key-frame is extremely slow. I'm pretty sure this is caused by the following
https://github.com/napari/napari-animation/blob/f5f797909e798c8da5c27cd72fdf7f3695bcb62c/napari_animation/animation.py#L142-L146

Which explicitly sets a bunch of properties on the layers, triggering redraws which can sometimes be expensive (e.g. larger 3D volumes)

I see two ways to solve this:
- checking for equality before setting properties to minimise redraws (easy, naive)
- attempting to only store diffs between key-frames as suggested by @sofroniewn in #40 

Will attempt the first option and if that doesn't make things feel smooth later will investigate more thoroughly

alisterburt:
naive solution works!

sofroniewn:
Ok nice, yeah - good catch - one we refactor the layers to be evented pydantic models we'll get the equality check for free, but till then we have to add it ourselves

sofroniewn:
Closed by #55

PR #55, opened 2021-04-25
Add equality check to avoid expensive redraws

alisterburt:
A simple equality check when setting layer properties solves #54 - we go back to instant keyframe switches!

We can think about hashing states if we notice performance issues down the line

@sofroniewn @guiwitz I think this should be added before release of 0.0.1 if you find time to review/test

edit: np.array_equal is used as it handles both normal and array comparisons


sofroniewn:
Ok, nice @alisterburt - I think we should merge this as is. As noted in the other issue, once we get the layer models to be evented models we'll get this for free - but till then best to do this ourselves

sofroniewn (approved):

Issue #50, opened 2021-04-23
changing keyframe in GUI with keyboard does not update viewer state


alisterburt:
Currently when we click on different key-frames in the GUI we set the viewer state to the selected key-frame - when moving up and down the key-frames list with the arrow keys the viewer does not update.

This is because the change is currently connected to the `itemClicked` event, need to change to `itemSelectionChanged` to fix

alisterburt:
closed by #57

Issue #51, opened 2021-04-23
AnimationSliderWidget does not update animation.frame


alisterburt:
Currently if I navigate through the animation with the new `AnimationSliderWidget` then add a new key-frame it gets added to the end of the list

I think the selected key-frame is being updated in the `KeyFramesListWidget` but `Animation.key_frame` is not staying in sync with this update - a fix for this will probably also solve #50

alisterburt:
closed by #71

PR #57, opened 2021-04-25
change from itemClicked to itemSelectionChanged in KeyFramesListWidget

alisterburt:
fix for #50 
may also solve #51


sofroniewn:
I can confirm this fixes #50, i might not quite be understanding #51, I couldn't reproduce it. Regardless this change looks good. I will merge now

PR #58, opened 2021-04-25
update frame number on drop in KeyFramesListWidget

alisterburt:
Previously, dropping a key-frame in the `KeyFramesListWidget` would not update the current key-frame index stored in the `Animation` - this is fixed in this PR


sofroniewn:
LGTM!

PR #60, opened 2021-04-25
add codecov config

alisterburt:
as suggested by @sofroniewn - should fix failures on #57 and #58 due to drop in code coverage


sofroniewn (approved):
LGTM! Thanks

Issue #56, opened 2021-04-25
Add pre-commit


alisterburt:
We should look at adding a pre-commit config to the repo

We can use the napari config as a starting point
https://github.com/napari/napari/blob/master/.pre-commit-config.yaml

I tried quickly myself a few weeks ago and black and flake8 were arguing with each other...

@Fifourche

alisterburt:
closed by #61

PR #61, opened 2021-04-26
Added a pre-commit config file, and ran it on the existing code.

Fifourche:
To tackle #56 , I added a config file for `pre-commit` that I've taken from `napari`. I modified it by commenting the `napari` metadata and configuring `isort` to be compatible with `black`.


alisterburt (approved):
thanks for this @Fifourche - this looks good to me. I'll wait for confirmation from @sofroniewn  before merging

sofroniewn:
Looking good - though can we now move all our stuff over from `setup.py` to `setup.cfg`? Maybe @tlambert03 can advise here too?

Also it looks like all our strings now use `"` instead of `'`. I guess that's ok, but i think in napari core we disabled that. I don't have strong feelings though.

@alisterburt, let's see if @tlambert03 weighs in, but then I'd say good for you to merge when you feel ready

Fifourche:
I moved almost everything from setup.py to setup.cfg (thanks for the info!!) ; I hope I did it right ! 

Regarding quotes, I checked different files on `napari` and it seems to be always `'` except when there are already some in the string, and they need to be "escaped" ; in this case `"` is used. Which makes sense to find also in here, as I pretty much took the config from `napari`! Anyways if it's to be changed I'll find a way to do it :)

setup.cfg
```diff
@@ -0,0 +1,75 @@
...
+url = https://github.com/napari/napari-animation
+download_url = https://github.com/napari/napari-animation
+license = BSD 3-Clause
+license_file = LICENSE
+description = A plugin to make animations with napari
+long_description = file: README.md
+long_description_content_type = text/markdown
+author = Nicholas Sofroniew
+author_email = sofroniewn@gmail.com
+classifiers=[
```
tlambert03, line 13:
```suggestion
classifiers =
```

setup.cfg
```diff
@@ -0,0 +1,75 @@
...
+name = napari_animation
+version = 0.0.1
+url = https://github.com/napari/napari-animation
+download_url = https://github.com/napari/napari-animation
+license = BSD 3-Clause
+license_file = LICENSE
+description = A plugin to make animations with napari
+long_description = file: README.md
+long_description_content_type = text/markdown
+author = Nicholas Sofroniew
```
tlambert03, line 11: plenty of people working on this now, could consider using comma separated names?

setup.cfg
```diff
@@ -0,0 +1,75 @@
+[metadata]
+name = napari_animation
+version = 0.0.1
```
tlambert03, line 3: if you're going to use `setuptools_scm` for version management (which is what is indicated by the `use_scm_version` bit in `setup.py`) ... then you can leave out the version here.  If we have it in the napari `setup.cfg`, it's for "odd" reasons :)

tlambert03 (commented):
setup stuff looks good!  you can test it locally by running `python setup.py --version` ... and it should give you an output that represents the most recent git tag (if all is well with setuptools_scm)

alisterburt:
Thanks for the input @tlambert03 !

@Fifourche - it looks like the string settings are controlled in the `pyproject.toml` file in napari
https://github.com/napari/napari/blob/master/pyproject.toml#L29-L53

These can probably also be set in setup.cfg, I'm not sure why the isort/black/flake8 settings are divided across `setup.cfg` and `pyproject.toml` in napari

tlambert03:
> These can probably also be set in setup.cfg, I'm not sure why the isort/black/flake8 settings are divided across setup.cfg and pyproject.toml in napari

this part is a bit of a mess.  each tool makes its own decisions on what it's going to support, and they're all over the board at the moment.  Black only supports [pyproject](https://black.readthedocs.io/en/stable/pyproject_toml.html), but some of the others don't support it at all...  So there's unfortunately no way yet to have only one file :/

I recommend this article if you're curious: https://snarky.ca/what-the-heck-is-pyproject-toml/

alisterburt:
Appreciate the links Talley! Looks like everything is moving in the right direction 🙂

setup.cfg
Fifourche: Yep ; I added the Github contributors names to the list, but I don't know what was expected (you're involved but your name isn't there). Are there guidelines for that ?

alisterburt:
Thanks @Fifourche for the fixes - I removed some napari specific black config (following guidelines from https://black.readthedocs.io/en/stable/pyproject_toml.html about string normalisation for new projects - given that the strings are now all with double quotes anyway I guess we should follow the black style) - let's merge!

alisterburt:
hmm, CI now failing on windows - was working 3 days ago with the same testing config in #60

sofroniewn:
Nice job @Fifourche, thanks for the help also @tlambert03 and great work @alisterburt shepherding this to the finish line!

Issue #37, opened 2021-03-05
reintroduce slerp for camera interpolation


alisterburt:
I see that camera angles were previously interpolated using spherical linear interpolation [here](https://github.com/sofroniewn/napari-animation/commit/0fb9af0c2d8626b3797facd14ca96179450bb227#diff-ddb53d972021c265e3350c22a11296b5338bf03014148ad91e692e550bdea6f1) but we've now moved to the type based interpolation which directly interpolates the three Euler angles.

Was this done to simplify the code when extending out to interpolating state, rather than just camera properties? If not, is there another reason I'm missing?

sofroniewn:
Yes, this was done to make interpolation as automated as possible, but there are sometime when we want a custom interpolation, for example we want zoom to be exponentially transformed and camera angle to be spherical linear interpolation. We should probably think of some general mechanism where we can specify some properties where we want custom transforms and then do type based interpolation.

Maybe we can provide a dict of dicts like 
```python
{
'camera.zoom':{'forward': lambda x: np.power(10, x), 'inverse':  np.log10)},
'camera.angles':{'forward': some_function, 'inverse': some_inverse_function},
}
```
and then transform each attribute accordingly. How does this sound?

Fifourche:
Hi ! I'm new to all this (napari-animation and github), but I had some questions related to this issue.

Would this Slerp be reintroduced directly on the ArcballCamera _quaternion attribute, or via the Camera Euler angles ? I guess that there would still be some issues  related to gimbal lock if we're using an Euler angles based view description, right ?

alisterburt:
Hi @Fifourche - thanks for your input!

I think we would likely want to stick to using the `angles` property of the camera as input for the state interpolation because this is what's exposed publicly to the user in napari (as `Viewer.camera.angles`). We get that state directly from the camera [here](https://github.com/napari/napari-animation/blob/main/napari_animation/animation.py#L87-L103).

There shouldn't be any problems with gimbal lock because the angles are [generated from/used to generate quaternions](https://github.com/napari/napari/blob/master/napari/_vispy/vispy_camera.py#L45-L70) for the camera. The underlying rotation matrix/quaternion representation generated from the Euler angles is still correct.

If you'd like to help us out on this but would benefit from help contributing do let me know, happy to help you get up and running!

alisterburt:
closed by #62

PR #62, opened 2021-04-27
Slerp + dict of interpolation functions

Fifourche:
This PR is to propose a way for the user to give custom interpolation functions to compute frames. It also contains a Slerp default method. A related issue has been opened at #37 .

## About custom interpolation functions
In details, I added a "methods" argument to the `capture_keyframe` function. "methods" is a dict of dicts containing functions, normally of same structure than "state". 

For instance, the custom interpolation function for the `state['camera']['angles']` would be looked for at `methods['camera']['angles']`.

A few things to notice and to be discussed I think :

- I made it so if an user wants to provided only a custom function for the "zoom" attribute for instance, he wouldnt' have to fill a whole dict of dicts ; he could just pass `methods = {'camera': {'zoom': custom_function}}`. Other attributes' interpolation functions would be defaulted.
- an interpolation function is expected to be of the form `custom_func(a, b, fraction)`
- it is possible to pass a function for a whole subpart of "state" ; i.e. if the interpolation should intertwine all "camera" attributes, we can pass `methods = {'camera': custom_func}`. 
`custom_func` will take `initial_state['camera']` and `final_state['camera']` dicts as `a` and `b` arguments

## About Slerp
I took advantage of this to add a Slerp method, relying on `scipy`. The default interpolation function for "angles" is set to Slerp.

### An example code :
Here's an example code where "methods" is either set to a standard linear interpolation for "angles" and or defaulted, resulting in Slerp for "angles".
Playing with the slider will help visualizing the difference.

```Python
import napari
from napari_animation import AnimationWidget
from skimage.data import cells3d

cells = cells3d()[:, 1]  # get some data

# Launch viewer to view data
viewer = napari.view_image(cells, colormap='magma')
viewer.dims.ndisplay = 3

# Instantiate the animation widget
animation_widget = AnimationWidget(viewer)
viewer.window.add_dock_widget(animation_widget, area='right')
animation = animation_widget.animation

# Angles interpolation is set to simple linear interpolation of a sequence
from napari_animation.utils import _interpolate_seq
methods_test = {'camera': {'angles': _interpolate_seq}}
# methods_test = {} # uncomment to default to Slerp instead

#%% Capture frames

animation.key_frames.clear()

viewer.camera.angles = (0, 0, 90)
animation.capture_keyframe(methods=methods_test)

viewer.camera.angles = (0, 0, 180)
animation.capture_keyframe(methods=methods_test)

viewer.camera.angles = (-180, -90, 0)
animation.capture_keyframe(methods=methods_test)

viewer.camera.angles = (180, 0, 0)
animation.capture_keyframe(methods=methods_test)

viewer.camera.angles = (-180, 90, 0)
animation.capture_keyframe(methods=methods_test)

viewer.camera.angles = (0, 0, -180)
animation.capture_keyframe(methods=methods_test)

viewer.camera.angles = (0.0, 0.0, 90.0)
animation.capture_keyframe(methods=methods_test)
```


alisterburt (commented):
Awesome PR @Fifourche - thanks so much for the time and effort!!

There are some things I like about this and some I like a bit less, I'd be really interested to hear your thoughts and what @sofroniewn and @guiwitz think about the following...

- I like providing a custom interpolation function with the signature `f(a, b, fraction)` as you suggested more than providing separate 'forward' and 'inverse' transforms prior to standard interpolation as suggested by @sofroniewn in #37. I think for more complicated transforms (like slerp and the relevant euler2quaternion stuff in this PR) it's easier to reason about and keep tidy... happy to be told I'm wrong here though!

- I'm not so sure on this being exposed as a keyword argument in the `Animation.capture_keyframe()`, I think it should be managed managed by us and invisible to the user. I can't imagine a situation where I want to manually set these to something different than the 'correct' function for a given attribute

- the cascading functionality is nice but I can't imagine a situation where I want one custom function for multiple attributes, did you have a specific use case in mind?

- I think I prefer the notation @sofroniewn used to define attributes `'camera.zoom'` etc
```Python
{
'camera.zoom':{'forward': lambda x: np.power(10, x), 'inverse':  np.log10)},
'camera.angles':{'forward': some_function, 'inverse': some_inverse_function},
}
```

---

Given what I've said above I think what we might want to do is
- create an `Interpolation` enum like we do for easing functions, in a new module called `interpolation.py`, this should contain all current and future interpolation functions, moving them out of `utils.py`

Easing func enum as an example is here
 https://github.com/napari/napari-animation/blob/829bc31cbd222453b08157d7a77d544974e21fc9/napari_animation/easing.py#L242-L289

- create a dict of form
```python
{
'camera.zoom' : Interpolation.LOG,
'camera.angles' : Interpolation.SLERP
}
```
for attributes which require a custom interpolation function

- add a mechanism for parsing that dict and using the specific functions on the specified attribute

We could also drop the enum and simple have a dict of functions - curious to hear your thoughts!!

sofroniewn:
I'm afraid I havn't got too much time right now to think deeply about this, but I like the idea of getting rid of my `pow 10` and `log` functions for zoom etc. that really did feel like a hack. I also tend to agree with a lot of @alisterburt points about API, keeping this hidden from users as much as possible if we know what should be done for an attribute, and using an API something like 
```python
{
'camera.zoom' : Interpolation.LOG,
'camera.angles' : Interpolation.SLERP
}
```
which seems quite elegant.

I think this PR might also contain some changes from #61, so once we get that merged this should be easier to review. I should be able to get to it early next week, but in the mean time I say press on! 🚀

Fifourche:
Thanks for your thoughts ! :) I agree with you, it should probably stay invisible to the user !

For this and for the cascading, what I had in mind at the time was the interpolation of the recently added layers' states. Some of the properties -not the base ones, but things like colormap or others- might be tricky to interpolate, and I imagined scenarios where the interpolation fucntion would need to change opacity as well as other parameters for instance.

It was and still isn't clear in my mind how to prepare for this, but in any case I agree with everything you said above, and will adopt the proposed form ! Thanks again 😃

alisterburt:
awesome - look forward to seeing what you come up with @Fifourche !

You should join us on the [napari zulip chat](https://napari.zulipchat.com/) - there are also developer meetings each week, it would be great to have you there!

Fifourche:
I made some changed according to the discussion !

I created a dictionary :
```Python
interpolation_dict = {
    "camera.angles": Interpolation.SLERP,
    "camera.zoom": Interpolation.LOG,
}
```
The unspecified properties will be defaulted to `Interpolation.LINEAR`. One thing I don't know is where to place this dict ; I put it at the very end of `interpolation.py` but it may be a bit hidden there.

Also, I guess some tests should be changed accordingly ; I'll take care of that soon if you're ok with the commits so far !

napari_animation/interpolation.py
```diff
@@ -0,0 +1,91 @@
...
+from functools import partial
+
+import numpy as np
+from napari._vispy.quaternion import quaternion2euler
+from scipy.spatial.transform import Rotation as R
+from scipy.spatial.transform import Slerp
+from vispy.util.quaternion import Quaternion
+
+
+def default(a, b, fraction):
```
sofroniewn, line 12: Can you add `Parameters` and `Returns` sections to these doc strings

napari_animation/interpolation.py
```diff
@@ -0,0 +1,91 @@
+import numbers
+from enum import Enum
+from functools import partial
+
+import numpy as np
+from napari._vispy.quaternion import quaternion2euler
+from scipy.spatial.transform import Rotation as R
+from scipy.spatial.transform import Slerp
+from vispy.util.quaternion import Quaternion
```
sofroniewn, line 9: I guess this could be ok for now, but I'd prefer if we didn't need the direct vispy import here. Maybe there is quaternion support inside scipy, or we can just vendor/ define the functions/ objects we need in a utils file here.

The reason is one day napari might support other rendering engines and so I'd like to keep our coupling to vispy as light as possible. This seems like one place where we could have a vispy independent alternative

napari_animation/interpolation.py
```diff
@@ -0,0 +1,91 @@
...
+
+    def __call__(self, *args):
+        return self.value(*args)
+
+
+# Dictionary relating state attribtues to interpolation functions
+interpolation_dict = {
+    "camera.angles": Interpolation.SLERP,
+    "camera.zoom": Interpolation.LOG,
+}
```
sofroniewn, line 91: I wonder if we shouldn't move these to a different file/ locations, maybe put them directly onto our animation object. They could either be private, or I could imagine someone might want to overwrite them, add their own custom attributes to this dictionary and having them on `Animation` seems like a good place.

napari_animation/utils.py
```diff
@@ -13,55 +15,62 @@ def interpolate_state(initial_state, final_state, fraction):
...
+
+
+def _easing_func_to_name(easing_function):
+    [name] = [e.name for e in Easing if e.value is easing_function]
+    return name
 
 
-def _interpolate_int(a, b, fraction):
-    return int(_interpolate_float(a, b, fraction))
+def nested_get(input_dict, keys):
```
sofroniewn, line 50: Can you add `Parameters` and `Returns` sections to the doc strings here and to the other nested functions

napari_animation/utils.py
```diff
@@ -1,8 +1,10 @@
 from .easing import Easing
+from .interpolation import Interpolation, interpolation_dict
 
 
 def interpolate_state(initial_state, final_state, fraction):
```
sofroniewn, line 5: This might make sense to put inside the `interpolation.py` file now, but I don't really mind either way

sofroniewn (approved):
Looking really great! I've left mostly minor, cosmetic comments, flagged a couple doc strings that need a few more sections, but otherwise this looks great and I think will then be good to merge @alisterburt when you're ready.

I'd also like to second @alisterburt comment about the [napari zulip](https://napari.zulipchat.com/login/) and communities meetings. I see you're in the zulip now, so I'll reach other there and give you additional details on the meetings.

napari_animation/interpolation.py
Fifourche: I added some to all these functions, it might be silly but just in case. ^^

napari_animation/interpolation.py
Fifourche: For now, I left it in `interpolation.py` ! I agree that we might want to define it as an `Animation` attribute and pass it as a parameter to `interpolate_state`, if there's no objection !

sofroniewn: @alisterburt what do you think? Agreed it's something that can come later, so we can merge this first and see

napari_animation/interpolation.py
sofroniewn: No it's really great! They should all have doc strings. We'll start building docs and sharing them online soon and then it will be really nice to have complete coverage in the standard form. Thanks!!

sofroniewn:
@Fifourche we've got some merge conflicts now since I merged #68, do you mind updating this PR. Then I think it is good to go!!

napari_animation/interpolation.py
Fifourche: I'm ok with `state_interpolation_map`, and added it ! Should be ok now :)

alisterburt: you're on fire! Will have a quick play then get to merging

alisterburt (approved):
Merge time 🚀

PR #63, opened 2021-04-29
add code style checks to github actions

alisterburt:
This PR adds the code style checks from napari's github actions into this repo, #61 should merge before we merge this


alisterburt:
#61 now merged and tox.ini fixed - ready for review

guiwitz:
I'm really not super familiar with setting up these code formatings, so ignore if this is not relevant, but if we add those checks with gh-actions upon committing, shouldn't we also have the whole machinery installed and documented when people do a developer installation? I don't think that at the moment pre-commits, black etc. get automatically installed like in the ```dev``` [section](https://github.com/napari/napari/blob/f49e81a15c024a0a67124907f5be70e9185e5eaa/setup.cfg#L106) of the ```setup.cfg``` of napari. Also this should be explained in the contributing section of the Readme as it adds a non-trivial amount of complexity for casual contributors.

alisterburt:
great feedback @guiwitz - you're totally right - I'll mark this as a draft and add some documentation in the readme

sofroniewn (approved):
LTGM! But agreed with @guiwitz we should make our contributing guidelines contain all the info needed for people to get setup. You can take the contributing guidelines from napari https://github.com/napari/napari/blob/master/docs/developers/contributing.md and adapt them here @alisterburt

sofroniewn:
@alisterburt just following up here, seems like all it needs are some docs and then it's good to go.

I wonder though @tlambert03 if this is something we should also make an option from the cookiecutter though, thoughts?

tlambert03:
i'm ok with it.

alisterburt:
thanks @sofroniewn - have added some minimal info about pre-commit to the README while we wait for docs to mature in #80 , will merge now

PR #65, opened 2021-04-29
Selection callback fix

alisterburt:
This PR fixes a bug which was an unintended side effect of #57

 `Animation.frame` would not be in sync with the selected frame in the `KeyFramesListWidget`, this meant modifying the `FrameWidget` would modify the wrong key-frame (or raise an `IndexError`)

Need to add Qt tests to check for this kind of error (#29)


sofroniewn (approved):
LGTM! Good to flag that once we get Qt tests we should add one here too. Till then it can wait though

PR #68, opened 2021-04-30
Fix Easing

### LINKED ISSUE #67: Easing problem
Something that happenned during the last merges I guess.

If we capture a key frame from script without specifying an easing function, the key frame 'easing' value contains `None`. 
But then when we slide through the frames, there's a callback to display the right easing in the easing list, and it cannot find the proper easing name !

Everything is ok though if we add the key frames from the animation widget, as it passes an easing function to the `capture_keyframe function`.

### Some code to reproduce this :
```Python
import napari
from napari_animation import AnimationWidget
from skimage.data import cells3d

cells = cells3d()[:, 1]  # get some data

# Launch viewer to view data
viewer = napari.view_image(cells, colormap='magma')

# Instantiate the animation widget
animation_widget = AnimationWidget(viewer)
viewer.window.add_dock_widget(animation_widget, area='right')
animation = animation_widget.animation

viewer.dims.ndisplay = 3
viewer.camera.angles = (0.0, 0.0, 90.0)
animation.capture_keyframe()
viewer.camera.zoom = 2.4
animation.capture_keyframe()
```

### Error returns : 
```
  File "..\napari_animation\utils.py", line 66, in _easing_func_to_name
    [name] = [e.name for e in Easing if e.value is easing_function]

ValueError: not enough values to unpack (expected 1, got 0)
```

Fifourche:
This makes Easing functions directly callable, for instance `Easing.LINEAR(a, b, fraction)` instead of `Easing.LINEAR.value(a, b, fraction)`.

Also, the `capture_keyframe` function now default the ease to `Easing.LINEAR` instead of `None`. It makes things more coherent overall and fixes #67.

### DIFF
--- a/napari_animation/_qt/frame_widget.py
+++ b/napari_animation/_qt/frame_widget.py
@@ -30,7 +30,7 @@ class FrameWidget(QWidget):
 
     def get_easing_func(self):
         easing_name = str(self.easeComboBox.currentText())
-        easing_func = Easing[easing_name.upper()].value
+        easing_func = Easing[easing_name.upper()]
         return easing_func
 
     def update_from_animation(self):
--- a/napari_animation/_tests/test_animation.py
+++ b/napari_animation/_tests/test_animation.py
@@ -31,6 +31,7 @@ def test_capture_key_frame(empty_animation):
     animation.capture_keyframe()
     assert len(animation.key_frames) == 1
     assert "viewer" in animation.key_frames[0].keys()
+    assert animation.key_frames[0]["ease"]
 
 
 def test_set_to_key_frame(animation_with_key_frames):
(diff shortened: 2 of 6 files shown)


alisterburt (approved):
Thank you @Fifourche ! Great catch with #67 and I like the fix... if we end up with multiple Enums for functions we might want to make this a FunctionEnum which aso handles wrapping the functions up as partials but for now this looks perfect.

I'll wait for @sofroniewn before merge but looks good to me!

sofroniewn (approved):
Very nice @Fifourche! Looks good to me. Maybe you can add a test before merge too though, can be simple just so as to prevent something like #67 slipping through again.

Issue #66, opened 2021-04-29
Strange visual effects when scrolling AnimationSliderWidget


alisterburt:
There are strange visual effects when changing the `AnimationSliderWidget`

The cause is that the selected frame changes in the `KeyFramesListWidget` when the slider moves, triggering the selection callback which sets the animation to the current key-frame rather than the correct interpolated frame

This should be easy to fix by blocking that particular event which triggers that callback when the selection is changed by moving the slider

sofroniewn:
Close by #69

PR #69, opened 2021-04-30
avoid triggering selection callback when moving AnimationSliderWidget

alisterburt:
fix for #66 

Updating the item selected in the `KeyFramesListWidget` when moving the `AnimationSliderWidget` had the unintended side effect of triggering the selection callback which set the key-frame - blocking signals temporarily fixes this 

before:
https://user-images.githubusercontent.com/7307488/116756299-c748e880-aa03-11eb-92f9-3e49d19ba795.mp4

after:
https://user-images.githubusercontent.com/7307488/116756303-c9ab4280-aa03-11eb-89a1-1c81421ae6b6.mp4


sofroniewn (approved):
Great catch! Love the videos too, so clear to see what was wrong before and how it's fixed. Will get this in now!!

PR #71, opened 2021-05-03
update animation.frame on animationslider move

alisterburt:
fix for #51 - includes moving some logic in the callback into an update method on the `AnimationSliderWidget`


sofroniewn (approved):
LGTM! Let's wait until after #62 to merge in case there are merge conflicts.

PR #72, opened 2021-05-03
Fix bug when removing item from KeyFramesListWidget

alisterburt:
bugfix for the following bug

Capturing three key-frames then removing the second would raise an error - this error came from the selection callback being called in the middle of the `KeyFramesListWidget.takeItem()` call

This is the same source of error as #69 - have generalised the fix from that to make it accessible if it comes up again


sofroniewn (approved):
Ok nice, LGTM! Let's merge after tests pass

PR #74, opened 2021-05-04
Very minor fix ; frame type and delete callback

Fifourche:
Just a very minor fix ; the delete button callback wasn't working just after playing with the slider.

In cause, the frame attribute value was casted to a numpy type, and couldn't be used anymore for popping a frame out as in the delete button callback !


alisterburt (approved):
Looks good to me! Thanks @Fifourche

PR #75, opened 2021-05-05
Fix bug in setting the animation writer options

guiwitz:
Two types of ```imageio``` writers are used depending on the chosen file type. A series of file types (mov, avi etc) gives access to more options such as e.g. ```quality```. The problem is that the dot sign in the list of extensions used to check the desired format is missing. As the ```suffix``` parameter of a ```pathlib.Path``` object returns extensions *with* a dot, the more complex writer is just ignored. This is a silent bug as no error is produced and passed options like ```quality``` are just ignored. This PR fixes this problem.


sofroniewn (approved):
Nice, LGTM!

PR #83, opened 2021-05-12
Stylesheet fix

alisterburt:
fix broken plugin import due to deprecation of `qt_viewer.raw_stylesheet` in napari 0.4.8 napari/napari#2643 - fixed by mirroring the napari console implementation
https://github.com/napari/napari-console/blob/fd3a291331e052e591042850b56fd1db5815f784/napari_console/qt_console.py#L151-L171

### DIFF
--- a/napari_animation/_hookimpls.py
+++ b/napari_animation/_hookimpls.py
@@ -5,4 +5,8 @@ from ._qt import AnimationWidget
 
 @napari_hook_implementation
 def napari_experimental_provide_dock_widget():
-    return (AnimationWidget, {"name": "Wizard"})
+    widget_options = {
+        "name": "Wizard",
+        "add_vertical_stretch": False,
+    }
+    return AnimationWidget, widget_options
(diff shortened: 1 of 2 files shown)


alisterburt:
removing extra commit!!

sofroniewn:
This PR also has the skeleton docs, can you remove them, thanks!!

sofroniewn:
Do we need to bump a minimum napari version? or was this in `0.4.5` already? We might need to bump for other reasons, but if not lets leave it

sofroniewn:
Looks like it should have been in 0.4.5 https://github.com/napari/napari/pull/2241

alisterburt:
tested locally and working with 0.4.5! Think we're good to go

tlambert03:
hang on before merging this...  the console repo is not a good one to mimic here... since it needs to do specific things with theming because of upstream QtConsole weirdness.  in almost all cases, the best approach (if possible) is not to set the theme and stylesheet at all, but let it be naturally inherited from the parent.  lemme see if I can try

alisterburt:
good to know, thanks - I think the original theming was done before `provide_dock_widget` so indeed might be unnecessary now!

sofroniewn:
Ok, nice looks like this was all we needed, and we've also fixed the stretch. Will merge now. Thanks all!!

tlambert03:
nooo :joy: ... ok will submit a new PR :)

sofroniewn:
loooooolz

sofroniewn:
> hang on before merging this...

ahhhhhhhh

PR #84, opened 2021-05-12
Stylesheet fix take 2

tlambert03:
a little modification to the stylesheet strategy...

The only reason it was necessary to do all the manual theming in `napari-console`  is because `qt_console`, rather frustratingly, uses something _other_ than Qt's `setStylesheet` to set their stylesheet.  (see that they have a `self.style_sheet` attribute that we set [here](https://github.com/napari/napari-console/blob/fd3a291331e052e591042850b56fd1db5815f784/napari_console/qt_console.py#L162)) ... 
In all other cases, the widget will inherit the stylesheet of its parent.  So `AnimationWidget` will just naturally "look right" when docked in napari, no additional effort required.

`KeyFramesListWidget` does still need to update theme for now (but I'll submit another PR sometime that may simplify all that too)... here, I just made the stylesheet template look a little bit more like a stylesheet.  hope that's ok

### DIFF
--- a/napari_animation/_qt/animation_widget.py
+++ b/napari_animation/_qt/animation_widget.py
@@ -47,9 +47,6 @@ class AnimationWidget(QWidget):
         self._add_keybind_callbacks()
         self._add_callbacks()
 
-        # Update theme
-        self._update_theme()
-
     def _init_ui(self):
         """Initialise user interface"""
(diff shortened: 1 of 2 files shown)


alisterburt:
looks really great, especially the nice clean style sheet, thanks @tlambert03 !

Issue #81, opened 2021-05-12
Slerp does not accept interpolation fraction outside the range [0, 1]


alisterburt:
Some of our easing functions extend beyond the range 0-1 (e.g. bounce) - this causes an error when interpolating the camera angles with slerp

PR #85, opened 2021-05-13
Made a modified copy of scipy's slerp to handle spherical extrapolation.

Fifourche:
To solve #81, I took the Slerp class from scipy and allowed extrapolation. I put the code in "utils" to prevent "interpolation.py" to be even less readable, but I can move it !


napari_animation/utils.py
```diff
@@ -125,3 +126,99 @@ def quaternion2euler(quaternion, degrees=False):
...
+        compute_times = np.atleast_1d(compute_times)
+
+        # side = 'left' (default) excludes t_min.
+        ind = np.searchsorted(self.times, compute_times) - 1
+        # Include t_min. Without this step, index for t_min equals -1
+        ind[compute_times == self.times[0]] = 0
+        # if np.any(np.logical_or(ind < 0, ind > len(self.rotations) - 1)):
+        #     raise ValueError("Interpolation times must be within the range "
+        #                      "[{}, {}], both inclusive.".format(
+        #                         self.times[0], self.times[-1]))
```
alisterburt, line 213: am I right in saying that commenting these lines is the only difference between this and `scipy.spatial.transform.Slerp`?

alisterburt (commented):
@Fifourche awesome - thanks for the fix! 

If we need to 'vendor' the Slerp class it may be simpler to write (and maintain?) our own slerp func without a lot of the extra fluff this class provides (multiple 'times' in the call, input sanitisation) - I'll PR your PR with an alternative, let me know what you think?

For reference, the pattern in napari for 'vendoring' code is to place it in a specific 'vendored' folder and make the minimum changes required to that file. This allows easier diffing/PR-ing to the upstream code. (probably not necessary for a numerical operation like this, useful for more complex objects e.g. the vispy `VolumeVisual` in napari https://github.com/napari/napari/blob/master/napari/_vispy/vendored/volume.py)

alisterburt:
PR proposed at Fifourche/napari-animation#2

napari_animation/utils.py
Fifourche: Yes indeed ! ^^ I completely agree with you, I just thought 'vendored' was for pieces of code that were not modified and there just to avoid dependencies ! I'll go chek your PR !

Fifourche:
Cool ! :) I had switch to scipy `Rotation` for conversion between Euler angles and rotation but forgot to use it too for the back conversion ; @alisterburt 's PR then also get rid of `quaternion2euler` which was also 'vendored' from `napari` anyways ! Thanks a lot 🎉

alisterburt:
Awesome - great work @Fifourche, was fun to work on this together!!

PR #86, opened 2021-05-13
Set a range to the steps QSpinBox

Fifourche:
The QSpinBox for steps value from the widget was limited to a range of 0 to 99. I set it to a more coherent value of 1 to 100000. I arbitrarily chose this last value, corresponding to the biggest power often that fitted in the box while the widget was reduced to a minimal size.


alisterburt (approved):
Looks good!

PR #88, opened 2021-05-18
Add `KeyFrame` object

tlambert03:
Haven't been around for the earlier conversations ... so not sure how you'll feel about this, but I was starting to look into updating the `KeyFramesListWidget` to take advantage of some of the new `QListWidget` stuff in napari (since you're already using the `EventedList`) ... however, capturing a keyframe as a pure dict presents some challenges (particularly when it comes a set-based selection model), since dicts are not hashable.

Also, as a relative newcomer to the codebase, I found myself wondering "what exactly _is_ this mysterious keyframe object?"  What are the possible, keys, etc...

So this PR introduces a `KeyFrame` and `ViewerState` dataclass, and pulls some of the methods off of `Animation` that are just there to create a new KeyFrame from the current viewer and adds them as `KeyFrame.from_viewer` and `ViewerState.from_viewer`.

if you were liking the "keyframe-as-pure-dict" model, feel free to reject this... but something like this will be necessary if you want to use the `SelectableEventedList` and associated listviews.

### DIFF
--- a/napari_animation/key_frame.py
+++ b/napari_animation/key_frame.py
@@ -1,0 +1,78 @@
+from __future__ import annotations
+
+from dataclasses import dataclass
+from typing import TYPE_CHECKING
+
+import numpy as np
+
+from .easing import Easing
+from .utils import make_thumbnail
+
+if TYPE_CHECKING:
+    from napari import Viewer
+
+
+@dataclass
+class ViewerState:
+    """The state of the viewer camera, dims, and layers.
+
+    Parameters
+    ----------
+    camera : dict
+        The state of the `napari.components.Camera` in the viewer.
+    dims : dict
+        The state of the `napari.components.Dims` in the viewer.
+    layers : dict
+        A map of layer.name -> _base_state for each layer in the viewer
+        (excluding metadata).
+    """
+
+    camera: dict
+    dims: dict
+    layers: dict
+
+    @classmethod
+    def from_viewer(cls, viewer: Viewer):
+        """Create a ViewerState from a viewer instance."""
+        layers = {
+            layer.name: layer._get_base_state() for layer in viewer.layers
+        }
+        for d in layers.values():
+            d.pop("metadata")
+        return cls(
+            camera=viewer.camera.dict(), dims=viewer.dims.dict(), layers=layers
+        )
+
+
+@dataclass
+class KeyFrame:
+    """A single keyframe in the animation.
+
+    Parameters
+    ----------
+    viewer_state : ViewerState
+        The state of the viewer at this keyframe.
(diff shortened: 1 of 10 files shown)


napari_animation/utils.py
```diff
@@ -67,3 +69,82 @@ def keys_to_list(input_dict):
...
+            2 * (q[3] * q[0] + q[1] * q[2]),
+            1 - 2 * (q[0] * q[0] + q[1] * q[1]),
+        )
+
+    angles = (theta_1, theta_2, theta_3)
+
+    if degrees:
+        return tuple(np.degrees(angles))
+    else:
+        return angles
```
alisterburt, line 127: looks like this came in from a merge - have removed in tlambert-forks/napari-animation#1

napari_animation/key_frame.py
```diff
@@ -0,0 +1,78 @@
...
+
+    @classmethod
+    def from_viewer(cls, viewer: Viewer, steps=15, ease=Easing.LINEAR):
+        """Create a KeyFrame from a viewer instance."""
+        return cls(
+            viewer_state=ViewerState.from_viewer(viewer),
+            thumbnail=make_thumbnail(viewer.screenshot(canvas_only=True)),
+            steps=steps,
+            ease=ease,
+        )
```
alisterburt, line 78: LOVE all of this - so easy to reason about!

napari_animation/utils.py
```diff
@@ -67,3 +69,82 @@ def keys_to_list(input_dict):
...
+    thumbnail = convert_to_uint8(thumbnail)
+
+    # blend thumbnail with opaque black background
+    background = np.zeros(shape, dtype=np.uint8)
+    background[..., 3] = 255
+
+    f_dest = thumbnail[..., 3][..., None] / 255
+    f_source = 1 - f_dest
+    thumbnail = thumbnail * f_dest + background * f_source
+    return thumbnail.astype(np.uint8)
```
alisterburt, line 150: much clearer!

alisterburt (approved):
this is amazing @tlambert03 - thanks for taking the time, it makes everything so much nicer!

I've left a PR at tlambert-forks/napari-animation#1 to remove `quaternion2euler` which is no longer needed, will merge this when that's fixed!

tlambert03:
Thanks for fixing my merge snafu :)

Glad you like this.  Just want to alert you to the one thing that I think might want further attention: which is that the interpolation functions expect dicts, so I'm using asdict here.  But if you don't intend to support interpolation between arbitrary dicts (that is, if it's always going to be a key frame or viewer state), then we could update those functions too... in a later PR

alisterburt:
good point - will open an issue now!

PR #90, opened 2021-05-20
Switch key_frames to a SelectableEventedList

Fifourche:
This PR is a follow up on the #87 ! Starting from @tlambert03 suggestions, I'm trying to switch `Animation.key_frames` to a `SelectableEventedList`, and change the widgets accordingly.

This PR is functional and pass tests, but there is still a lot to discuss. Main modifications are :

- Switched key_frames to a `SelectableEventedList`
- Decided to mainly use the `key_frames.selection._current` reference (as opposed to the `active` one) to manage selected key frame, and linked it to `Animation.current_key_frame`. Still I defined an `active_key_frame` attribute in case.
- I unfroze KeyFrame, as we want to modify some of its contents (from the GUI via the steps widget for instance).
- Added an `__eq__` to `KeyFrame` as value equality is checked during some `SelectableEventedList` methods. Also added one on `ViewerState`.
- I modified as little as possible the KeyFramesListWidget as suggested in #87 
- Tried to manage callbacks there, so it's functional but I guess it's still way unoptimal.
- I added an update of displayed steps and ease widgets when the slider's moving.

I'm happy to start a discussion on this ! 😅

### DIFF
--- a/napari_animation/_qt/animation_widget.py
+++ b/napari_animation/_qt/animation_widget.py
@@ -158,8 +158,10 @@ class AnimationWidget(QWidget):
     def _delete_keyframe_callback(self, event=None):
         """Delete current key-frame"""
         if len(self.animation.key_frames) > 0:
-            self.animation.key_frames.pop(self.animation.frame)
+            self.animation.key_frames.pop(self.animation.current_key_frame)
+            print(1)
         if len(self.animation.key_frames) == 0:
+            print(2)
             self.keyframesListControlWidget.deleteButton.setEnabled(False)
             self.keyframesListWidget.setEnabled(False)
             self.frameWidget.setEnabled(False)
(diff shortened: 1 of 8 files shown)


napari_animation/_qt/animation_widget.py
```diff
@@ -167,17 +168,18 @@ def _delete_keyframe_callback(self, event=None):
...
     def _key_back_frame(self, event=None):
         """Go backwards in key-frame list"""
 
-        new_frame = (self.animation.frame - 1) % len(self.animation.key_frames)
-        self.animation.set_to_keyframe(new_frame)
-        self.keyframesListWidget.setCurrentRow(new_frame)
+        new_frame = (self.animation.current_key_frame - 1) % len(
+            self.animation.key_frames
+        )
+        self.animation.current_key_frame = new_frame
```
tlambert03, line 182: could use `self.animation.key_frames.select_previous()`

napari_animation/animation.py
```diff
@@ -159,7 +169,27 @@ def _frame_generator(self, canvas_only=True):
...
+    def active_key_frame(self):
+        try:
+            return self.key_frames.index(self.key_frames.selection.active)
+        except ValueError:
+            warnings.warn("No active frame selected !")
+            return None
+
+    @active_key_frame.setter
+    def active_key_frame(self, frame_index):
+        self.key_frames.selection.active = self.key_frames[frame_index]
```
tlambert03, line 192: as mentioned in my comment, I might make these return the objects themselves, rather than their indices.  I might also remove the warning and just return `None` (seems like a very reasonable thing just to have no active/current keyframe).  Finally, unless you have an immediate need for the `@setter`, you could punt on exposing that for now... reducing this all to:

```suggestion
    @property
    def current_key_frame(self):
        return self.key_frames.selection._current

    @property
    def active_key_frame(self):
        return self.key_frames.selection.active
```

Fifourche:
Wow, thanks a lot ! ^^ I agree with everything you said ! It's true that it's easier indeed to pass the current `KeyFrame` around rather than the playing with its index, but I thought we might want to keep an attribute with a similar use as `frame` had, i.e. a way for the user to set the current key_frame programmatically by giving an index ! I agree this should be decoupled from the way to pass current KeyFrame internally.

Also, I'm not sure to understand what would be the use case of having `current` AND `active` ; wouldn't it be simpler to just have one of both ?

I'll sleep on it, and I'll check #91 !

tlambert03:
> I thought we might want to keep an attribute with a similar use as frame had, i.e. a way for the user to set the current key_frame programmatically by giving an index !

sure, you could expose that... but it's not hard for them to just use `active = key_frames[n]`.  In napari, for layers if they wanted to select the `nth` layer, I think we'd just tell them to use `layers.selection.active = layers[n]` (rather than creating a new API to do the indexing for them)


> Also, I'm not sure to understand what would be the use case of having current AND active ; wouldn't it be simpler to just have one of both ?

yep!  I agree. I would only use one (i only had them both in there to follow your pattern).  I think most cases, it will be `active` that makes the most sense... so maybe use that until you run into limitations.

Fifourche:
So should I try to fix the codecov related problems before merging this one ? #91 should fix them anyways while changing for some nicer things...

alisterburt:
awesome work @Fifourche and thanks for reviewing @tlambert03 !!

I'm going to merge this as is because Talley's followup #91 contains his suggested changes, no worries for the codecov! Great job

PR #94, opened 2021-05-28
Update README badges

Fifourche:
Update some badges, though I still get a CSP error for the tests badge !


alisterburt (approved):
looks good!

PR #96, opened 2021-05-28
remove unnecessary duplication of state interpolation

alisterburt:
state interpolation code was duplicated - removed and fixed test to account for the change

took the opportunity to move some testing utilities into `util`, they could also be moved to `utils.testing` if people prefer

A PR will follow based on this fixing the messy asdict stuff

@Fifourche

### DIFF
--- a/napari_animation/_tests/test_interpolation.py
+++ b/napari_animation/_tests/test_interpolation.py
@@ -1,4 +1,3 @@
-import numbers
 from dataclasses import asdict
 
 import numpy as np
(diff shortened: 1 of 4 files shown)


Fifourche (approved):
Looks all good to me ! :) 

It reminds me that we probably need to fix the interpolation test ; for now it ignores fractions that are not 0 or 1... I had done that at the time because I couldn't find a way to test it, apart from specifying an expected result !

alisterburt:
good point! do you want to make an issue for this?

Issue #89, opened 2021-05-18
modify interpolation functions to accept `ViewerState` objects directly


alisterburt:
after #88 we should modify the interpolation function so that it accepts `ViewerState` objects directly as suggested by @tlambert03 

This will remove the need to call `asdict` here and tidy things up a bit

https://github.com/napari/napari-animation/blob/8c7119e69933bcba8f0263d3cab966f373a7cc24/napari_animation/animation.py#L137-L138

alisterburt:
done!

PR #97, opened 2021-05-28
Make state interpolation work on ViewerState objects

alisterburt:
Interpolation of viewer states is currently based on the nested getters and setters which work with dictionaries, not `ViewerState` objects, meaning we have to first turn our viewerstate into a dictionary with `dataclasses.asdict()`

https://github.com/napari/napari-animation/blob/f38467593df65e224a0573b2e12f2b7cbcb8348c/napari_animation/frame_sequence.py#L134-L145

This PR (which includes #96) solves #89 by adding dict-like methods `__getitem__`, `keys()` and `get()`) on the `ViewerState` which allows us to treat it as a nested dictionary and interpolate as usual.

note: I tried to do this first by modifying the interpolation machinery, my attempts were less clean than this. I fully accept that this may be more complicated than necessary but would be really interested to hear what you guys think? @tlambert03 @Fifourche

### DIFF
--- a/napari_animation/_tests/test_frame_sequence.py
+++ b/napari_animation/_tests/test_frame_sequence.py
@@ -5,6 +5,7 @@ import pytest
 
 from napari_animation.frame_sequence import FrameSequence
 from napari_animation.key_frame import ViewerState
+from napari_animation.utils import nested_assert_close
 
 
 def test_frame_seq(frame_sequence: FrameSequence):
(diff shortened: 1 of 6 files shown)


tlambert03:
> adding dict-like methods __getitem__, keys() and get()) on the ViewerState which allows us to treat it as a nested dictionary and interpolate as usual.

hmmm... this actually feels a bit stranger to me than just using `asdict`.  I really haven't looked that deeply into the `nested_set`/`nested_get` thing, but if that part is pretty set in stone (and it requires a dict), then I'd say just stick with `asdict`, (which is the standard way to convert a dataclass into a dict), rather than making the dataclass itself have dictlike-characteristics

alisterburt:
totally fair, I had a feeling that might be the case 🙂 thanks for taking a look!

the nested get/set stuff comes from wanting an easy way to specify interpolation functions for specific attributes which may be arbitrarily deeply nested in the `ViewerState` - there's some discussion in an [old PR](https://github.com/napari/napari-animation/pull/62#pullrequestreview-646381507) if you're interested

Fifourche:
I agree that this is related to how `_interpolate_state` is built (nested set/get, and the InterpolationMap format). I guess there would be a good re-thinking to be done then !

PR #99, opened 2021-05-29
Synced slider with frame index rather than active keyframe

Fifourche:
A PR to link the `AnimationSlider` with the frame index being set.

Before it was synced with the change of `key_frames.selection.active`, and now with the signal of an `Emitter` : `frame_index`, added to `Animation`. This way, if we use `set_movie_frame_index`, it'll put the slider at the frame's position and not at the closest key frame.

For some context, the idea I had was to sync the slider with a frame_index signal to prepare the way for a "previewer" capability. The preview would be launched via a method of `Animation`, or a button next to the slider.

### DIFF
--- a/napari_animation/_qt/animation_widget.py
+++ b/napari_animation/_qt/animation_widget.py
@@ -96,7 +96,7 @@ class AnimationWidget(QWidget):
         keyframe_list.selection.events.active.connect(
             self._on_active_keyframe_changed
         )
-        self.animation.events._frame_index.connect(
+        self.animation._frames.events._current_index.connect(
             self._on_frame_index_changed
         )
 
--- a/napari_animation/_qt/animation_widget.py
+++ b/napari_animation/_qt/animation_widget.py
@@ -96,7 +96,9 @@ class AnimationWidget(QWidget):
         keyframe_list.selection.events.active.connect(
             self._on_active_keyframe_changed
         )
-        self.animation.events.frame_index.connect(self._on_frame_index_changed)
+        self.animation.events._set_frame_index.connect(
+            self._on_frame_index_changed
+        )
 
     def _input_state(self):
         """Get current state of input widgets as {key->value} parameters."""
--- a/napari_animation/_qt/animation_widget.py
+++ b/napari_animation/_qt/animation_widget.py
@@ -96,6 +96,7 @@ class AnimationWidget(QWidget):
         keyframe_list.selection.events.active.connect(
             self._on_active_keyframe_changed
         )
+        self.animation.events.frame_index.connect(self._on_frame_index_changed)
 
     def _input_state(self):
         """Get current state of input widgets as {key->value} parameters."""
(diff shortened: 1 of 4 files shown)


napari_animation/frame_sequence.py
```diff
@@ -57,16 +57,20 @@ def _rebuild_frame_index(self, event=None):
...
-        f = 0
-        for kf0, kf1 in pairwise(self._key_frames):
-            for s in range(kf1.steps):
-                fraction = s / kf1.steps
-                self._frame_index[f] = (kf0, kf1, fraction)
-                f += 1
+        else:
+            f = 0
+            if len(self._key_frames) == 1:
+                kf1 = self._key_frames[0]
```
alisterburt, line 67: I don't really understand the purpose of this little section, could you explain it?

alisterburt (changes requested):
hey @Fifourche - I love the idea but am not 100% on the implementation, especially the generation of the whole list to obtain the required index...

Could you also explain the extra stuff you added into the cache? 😄

napari_animation/_qt/animation_widget.py
```diff
@@ -126,27 +129,21 @@ def _on_keyframes_changed(self, event=None):
...
-            self.animationSlider.setValue(frame_index)
-            self.animationSlider.blockSignals(False)
+    def _on_frame_index_changed(self, event=None):
+        frame_index = event.value
+        self.animationSlider.blockSignals(True)
+        self.animationSlider.setValue(frame_index)
+        self.animationSlider.blockSignals(False)
 
+    def _on_active_keyframe_changed(self, event):
+        active_keyframe = event.value
```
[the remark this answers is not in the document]
Fifourche: Yes and yes ! will do :)

napari_animation/frame_sequence.py
Fifourche: Previously, if there was only one keyframe, it was not added as a frame and _frames was empty. I just added this here so _frames would be empty only when there are no keyframes at all.

alisterburt:
Okay cool! Maybe we could refactor this logic into a ‘has_frames’ properly rather than adding it here?

Fifourche:
Wasn't sure of what you meant, as it's mostly the number of key frames that is needed there. Would you still go for an attribute for this ?

> Okay cool! Maybe we could refactor this logic into a ‘has_frames’ properly rather than adding it here?
> […](#)
> On 1 Jun 2021, at 12:19, Fifourche ***@***.***> wrote: ﻿ @Fifourche commented on this pull request. In napari_animation/frame_sequence.py: > self.events.n_frames(value=len(self)) return - f = 0 - for kf0, kf1 in pairwise(self._key_frames): - for s in range(kf1.steps): - fraction = s / kf1.steps - self._frame_index[f] = (kf0, kf1, fraction) - f += 1 + else: + f = 0 + if len(self._key_frames) == 1: + kf1 = self._key_frames[0] Previously, if there was only one keyframe, it was not added as a frame and _frames was empty. I just added this here so _frames would be empty only when there are no keyframes at all. — You are receiving this because you commented. Reply to this email directly, view it on GitHub, or unsubscribe.

napari_animation/frame_sequence.py
```diff
@@ -57,16 +57,20 @@ def _rebuild_frame_index(self, event=None):
...
+            else:
+                for kf0, kf1 in pairwise(self._key_frames):
+                    for s in range(kf1.steps):
+                        fraction = s / kf1.steps
+                        self._frame_index[f] = (kf0, kf1, fraction)
+                        f += 1
```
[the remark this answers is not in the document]
Fifourche: I still kept an `elif`, for probably a better readability !

napari_animation/animation.py
```diff
@@ -208,6 +230,12 @@ def animate(
         if not save_as_folder:
             writer.close()
 
+    def _keyframe_frame_index(self, keyframe):
+        n_frames = len(self._frames)
+        kf1_list = [self._frames._frame_index[n][0] for n in range(n_frames)]
```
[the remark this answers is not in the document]
Fifourche: I'm switching back to something I wrote on a previous version. I hope it feels lighter ! :)

Fifourche:
Hey :)
I'm changing a bit this PR to switch the current frame index attribute from `animation` to `FrameSequence` ! In my eyes it makes `FrameSequence` a bit more like an virtual slider, which would be relevant to connect to the GUI slider, and thus prepare the ground for a play button with #100 .

napari_animation/animation.py
```diff
@@ -208,6 +215,15 @@ def animate(
         if not save_as_folder:
             writer.close()
 
+    def _keyframe_frame_index(self, keyframe_index):
+        """Gets the frame index of the keyframe corresponding to keyframe_index."""
+        # Get all steps leading to keyframe.
+        steps_to_keyframe = [
+            kf.steps for kf in self.key_frames[1 : keyframe_index + 1]
+        ]
```
alisterburt, line 223: this is looking much nicer! 🙂

napari_animation/frame_sequence.py
```diff
@@ -160,3 +170,17 @@ def iter_frames(
...
+
+    @property
+    def _current_index(self):
+        return self.__current_index
+
+    @_current_index.setter
+    def _current_index(self, frame_index):
+        if frame_index != self._frame_index:
+            self.__current_index = frame_index
+            self.events._current_index(value=frame_index)
```
alisterburt, line 186: I really like having this directly on the `FrameSequence`

alisterburt (approved):
hey @Fifourche - this is looking really nice now! Thanks for your patience with my reviewing here, let's get this in!

PR #100, opened 2021-05-31
Add a preview functionality

Fifourche:
Following #99 , added functions to enable a preview ability. This preview capability is compatible with the slider as it allows to still change frame interactively using it (see video snippet).

https://user-images.githubusercontent.com/22133982/120185709-d98c9100-c212-11eb-8354-721b46145198.mp4

### DIFF
--- a/napari_animation/_qt/animation_widget.py
+++ b/napari_animation/_qt/animation_widget.py
@@ -96,7 +96,9 @@ class AnimationWidget(QWidget):
         keyframe_list.selection.events.active.connect(
             self._on_active_keyframe_changed
         )
-        self.animation.events.frame_index.connect(self._on_frame_index_changed)
+        self.animation.events._set_frame_index.connect(
+            self._on_frame_index_changed
+        )
 
     def _input_state(self):
         """Get current state of input widgets as {key->value} parameters."""
--- a/napari_animation/_qt/animation_widget.py
+++ b/napari_animation/_qt/animation_widget.py
@@ -96,6 +96,7 @@ class AnimationWidget(QWidget):
         keyframe_list.selection.events.active.connect(
             self._on_active_keyframe_changed
         )
+        self.animation.events.frame_index.connect(self._on_frame_index_changed)
 
     def _input_state(self):
         """Get current state of input widgets as {key->value} parameters."""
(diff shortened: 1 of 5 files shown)


sofroniewn:
Hi @Fifourche, sorry just seeing this now. Maybe @alisterburt @tlambert03 and I can give this some review when we get the chance. Am I right in thinking this is sort of like having a "play button" for our preview slider? If so we might be able to re-use some of the play button code we have for our dims slider. I'll check out #99 too when i can as well. Nice work here!!!

Fifourche:
Hi ! :) Yes that was the idea indeed, but I also wanted to give the ability to have a play function even without the GUI. 

That's why I started #99 in fact ! I wanted to have a `current_frame` attribute much like the `dims.current_step`, so that I could reuse the `dims_slider` play function there. In this PR I already use something similar to what's happening in the dims slider play (`AnimationWorker`, and precautions to allow the slider to still be interactive). 

I still need to work on #99 though, and give the `current_frame` attribute to the `FrameSequence` rather than `Animation` for instance.

sofroniewn:
@Fifourche - do you want to reuse the "play button" from inside napari here

image

It wasn't quite clear to me how to turn the playing on and off based on the GUI

Fifourche:
Hey ! Yes, that's I want to do indeed ! But as we didn't have any equivalent of the `Dims` element, I wanted to wait for #99 to have something similar ; now I want to replace the slider we have for an equivalent of the `QtDimSliderWidget` from napari, maybe without some options like reverse play. I should probably put this PR as a draft one in fact, sorry ! ^^

sofroniewn:
> I should probably put this PR as a draft one in fact, sorry ! ^^

No worries, all good! Let us know if we can help!!

Fifourche:
I have a working version of it, ~ready for review I guess ! :) 

What I'd like to do still is to move the `QtPlayButton` from here to `superqt`, and use subclasses if needed in `napari` and `napari-animation`. Do you think this would be worth it @tlambert03 ?

Fifourche:
The tests failed here but for a `napari` related issue ?

sofroniewn:
> The tests failed here but for a napari related issue ?

Ok thanks for flagging! @alisterburt or I should be able to review this soon, thanks!!

Fifourche:
Once again, a `napari` related issue :

```
    File "/home/runner/work/napari-animation/napari-animation/.tox/py39-linux-pyside/lib/python3.9/site-packages/napari/_qt/utils.py", line 304, in remove_flash_animation
      widget.setGraphicsEffect(None)
  RuntimeError: Internal C++ object (QtWidgetOverlay) already deleted.
```

Otherwise it seems ok I think !

PR #101, opened 2021-05-31
Fixed deleteButton and easings support

Fifourche:
There was an issue with easings, that are currently not taken into account ; a small piece of code was missing.
Also, the delete button is enabled by default even when animation contains no keyframe upon initialization.

This PR fixes both !


alisterburt (approved):
great catch!!

Issue #76, opened 2021-05-06
Expose advanced options in save dialogue


alisterburt:
In our `Animation.animate()` method we have quite a few options 

https://github.com/napari/napari-animation/blob/a2be8d9a93221f4aedcc1d13891da43d8c2bd35c/napari_animation/animation.py#L221-L313

None of these options are exposed in the GUI which is currently just a save button and uses the options below

https://github.com/napari/napari-animation/blob/a2be8d9a93221f4aedcc1d13891da43d8c2bd35c/napari_animation/_qt/animation_widget.py#L200-L201

It might be a good idea to expose some of these options. To avoid cluttering up the GUI, I think these should be somehow available within the save dialog box

alisterburt:
closed by #102

Issue #39, opened 2021-03-05
save dialog issues


alisterburt:
I'm trying to incorporate extra layer metadata [over here](https://github.com/alisterburt/napari-animation/tree/layer-visibility) and keep running into what seems to be a low level Qt/system bug when trying to work with the save dialog. 

This has worked previously but the issue persisted after restart so leaving an issue here

```
2021-03-05 20:03:46.759 python[8310:261652] WARNING: <NSSavePanel: 0x7fecc3bfa3e0> -[NSSavePanel(exportedInterface) requestAppEnabledStateForItems:replyBlock:] no longer has a delegate to respond to shouldEnableURL
```

this [apple 'appkit' dev page](https://developer.apple.com/documentation/appkit/nssavepanel) has some info

sofroniewn:
Hmm, i havn't seen this one before. Is it still a problem for you?

alisterburt:
Has disappeared, thankfully, probably fixed in some OS update!

PR #102, opened 2021-06-02
Added a save dialog with options

Fifourche:
This PR aims at improving the save procedure, and brings a draft of save dialog with options, as raised by #76 .
Currently filters are not properly handled. The rest seems ok !

save_na

### DIFF
--- a/napari_animation/_qt/animation_widget.py
+++ b/napari_animation/_qt/animation_widget.py
@@ -4,7 +4,6 @@ from napari import Viewer
 from qtpy.QtCore import Qt
 from qtpy.QtWidgets import (
     QErrorMessage,
-    QFileDialog,
     QPushButton,
     QSlider,
     QVBoxLayout,
@@ -15,6 +14,7 @@ from ..animation import Animation
 from .frame_widget import FrameWidget
 from .keyframelistcontrol_widget import KeyFrameListControlWidget
 from .keyframeslist_widget import KeyFramesListWidget
+from .savedialog_widget import SaveDialogWidget
 
 
 class AnimationWidget(QWidget):
@@ -49,6 +49,7 @@ class AnimationWidget(QWidget):
         )
         self.frameWidget = FrameWidget(parent=self)
         self.saveButton = QPushButton("Save Animation", parent=self)
+        # self.saveDialogWidget = SaveDialogWidget(parent=self)
         self.animationSlider = QSlider(Qt.Horizontal, parent=self)
         self.animationSlider.setToolTip("Scroll through animation")
         self.animationSlider.setRange(0, len(self.animation._frames) - 1)
@@ -152,24 +153,38 @@ class AnimationWidget(QWidget):
 
     def _save_callback(self, event=None):
 
-        if len(self.animation.key_frames) < 2:
+        try:
+            self.animation._validate_animation()
+        except ValueError as err:
             error_dialog = QErrorMessage()
-            error_dialog.showMessage(
-                f"You need at least two key frames to generate \
-                an animation. Your only have {len(self.animation.key_frames)}"
-            )
+            error_dialog.showMessage(str(err))
             error_dialog.exec_()
+            return
 
-        else:
-            filters = (
-                "Video files (*.mp4 *.gif *.mov *.avi *.mpg *.mpeg *.mkv *.wmv)"
-                ";;Folder of PNGs (*)"  # sep filters with ";;"
+        filters = (
+            "Video files (*.mp4 *.gif *.mov *.avi *.mpg *.mpeg *.mkv *.wmv)"
+            ";;Folder of PNGs (*)"  # sep filters with ";;"
+        )
+
+        saveDialogWidget = SaveDialogWidget(self)
+        (
+            filename,
+            fps,
+            quality,
+            canvas_only,
+            scale_factor,
+        ) = saveDialogWidget.getSaveFileName(
+            self, "Save animation", str(Path.home()), filters
+        )
+
+        if filename:
+            self.animation.animate(
+                filename,
+                fps=fps,
+                quality=quality,
+                canvas_only=canvas_only,
+                scale_factor=scale_factor,
             )
-            filename, _filter = QFileDialog.getSaveFileName(
-                self, "Save animation", str(Path.home()), filters
-            )
-            if filename:
-                self.animation.animate(filename)
 
     def _nframes_changed(self, event):
         has_frames = bool(event.value)
--- a/napari_animation/_qt/animation_widget.py
+++ b/napari_animation/_qt/animation_widget.py
@@ -49,7 +49,6 @@ class AnimationWidget(QWidget):
         )
         self.frameWidget = FrameWidget(parent=self)
         self.saveButton = QPushButton("Save Animation", parent=self)
-        # self.saveDialogWidget = SaveDialogWidget(parent=self)
         self.animationSlider = QSlider(Qt.Horizontal, parent=self)
         self.animationSlider.setToolTip("Scroll through animation")
         self.animationSlider.setRange(0, len(self.animation._frames) - 1)
--- a/napari_animation/_qt/animation_widget.py
+++ b/napari_animation/_qt/animation_widget.py
@@ -120,11 +120,13 @@ class AnimationWidget(QWidget):
             raise ValueError("No selected keyframe to delete !")
 
     def _on_keyframes_changed(self, event=None):
+        n_keyframes = len(self.animation.key_frames)
         has_frames = bool(self.animation.key_frames)
 
         self.keyframesListControlWidget.deleteButton.setEnabled(has_frames)
         self.keyframesListWidget.setEnabled(has_frames)
         self.frameWidget.setEnabled(has_frames)
+        self.saveButton.setEnabled(n_keyframes > 1)
 
     def _on_active_keyframe_changed(self, event=None):
         n_frames = len(self.animation._frames)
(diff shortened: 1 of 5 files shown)


alisterburt:
hey @Fifourche, sorry for not following up on this sooner! I love the look of this from your screenshot but unfortunately can't play with it properly here, I keep running into the same issues as #39 (not your fault, I think this is a weird mac specific bug

alisterburt:
Following up here, what do you mean exactly by 'filters are not handled properly'? I've been meaning to fix this save dialog issue properly on my machine and getting this in is great motivation 🙂

Fifourche:
Hi ! :) I just wanted to say that the drop down menu is not displaying indivdually all the format options, just folder or else. Otherwise everything seems to work fine, even though tests would be needed !

I don't have access to a Mac, so I'm not sure I can help much... but you can try this to test the interface :
https://mybinder.org/v2/gh/Fifourche/omero-analysis-desktop.git/HEAD?filepath=napari.ipynb

This would run a `napari` viewer on a mybinder server, and let you play with it. Of course not optimized for real visualization, but might be helpful to check things in case !

setup.cfg
```diff
@@ -36,6 +36,7 @@ install_requires =
 	numpy
 	qtpy
 	scipy
+    tqdm>=4.56.0
```
alisterburt, line 39:
```suggestion
        tqdm>=4.56.0
```

napari_animation/_qt/savedialog_widget.py
```diff
@@ -0,0 +1,98 @@
...
+    QSpinBox,
+    QWidget,
+)
+
+
+class SaveDialogWidget(QFileDialog):
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+
+    def getSaveFileName(
```
alisterburt, line 18: this method now gets more than just the filename, should maybe be called `getAnimationParameters`

napari_animation/_qt/animation_widget.py
```diff
@@ -152,24 +154,36 @@ def _on_slider_moved(self, event=None):
...
-            if filename:
-                self.animation.animate(filename)
+        saveDialogWidget = SaveDialogWidget(self)
+        (
+            filename,
+            fps,
+            quality,
+            canvas_only,
+            scale_factor,
+        ) = saveDialogWidget.getSaveFileName(
```
alisterburt, line 169: this return would need to be modified if we change to the dict of kwargs

napari_animation/_qt/animation_widget.py
```diff
@@ -152,24 +154,36 @@ def _on_slider_moved(self, event=None):
...
+
+        if filename:
+            try:
+                self.animation.animate(
+                    filename,
+                    fps=fps,
+                    quality=quality,
+                    canvas_only=canvas_only,
+                    scale_factor=scale_factor,
+                )
```
alisterburt, line 181: same as above, would become 
```
self.animation.animate(**animation_kwargs)
```

napari_animation/_qt/savedialog_widget.py
```diff
@@ -0,0 +1,98 @@
...
+
+        # Get info back from user
+        if self.exec_():
+            filename = list(self.selectedFiles())[0]
+            fps = self.optionsWidget.fpsSpinBox.value()
+            quality = int(self.optionsWidget.qualityComboBox.currentText())
+            canvas_only = self.optionsWidget.canvasCheckBox.isChecked()
+            scale_factor = self.optionsWidget.scaleSpinBox.value()
+
+            return filename, fps, quality, canvas_only, scale_factor
```
alisterburt, line 55: it might be nicer if we return a dict of keyword arguments to `Animation.animate()`

something like

```python
animation_kwargs = {}
animation_kwargs['filename'] = list(self.selectedFiles())[0]
...etc

return animation_kwargs
```
```

**Review by alisterburt — APPROVED** [review]
@Fifourche you've really gone above and beyond providing that binder link! Thanks so much and sorry for not being able to review properly without it 🙂

I have a couple of small comments/suggestions but the functionality is awesome and works really nicely! Awesome job 

What do you think about making the `qualityComboBox` a `QLabeledSlider` from the new superqt repo? A slider might be more obvious than a combo box for the quality

I'm approving now but will wait to hear what you think about these suggestions before merging, if you'd rather not implement these changes yourself I'm happy to do them myself in a followup PR! 

Thanks again!!

File: setup.cfg [hunk 6]
```diff
@@ -36,7 +36,7 @@ install_requires =
 	numpy
 	qtpy
 	scipy
-    tqdm>=4.56.0
+        tqdm>=4.56.0
```
alisterburt, line 39:
```suggestion
      tqdm>=4.56.0
```

Fifourche:
Thanks for the suggestions, I took them all into account ! :) There's one thing I can't seem to work out for now ; it's the display style of the `QLabeledSlider`... I'm not sure how to handle the spacing properly, if you have some knowledge about this !

Fifourche:
Hey ! I'm still having problems with the looks of the `QLabeledSlider` for the quality ; the size of the label - the `QDoubleSpinBox`, seems to be set at a width way too large, as if it was taking into account up and down buttons even though they're not displayed...! Any advice for this, @tlambert03 ? :)
save_dialog

tlambert03:
hmm... mine actually looks good, what version of napari are you on?

Untitled

Fifourche:
Was still on 0.4.8, didn't realize... ^^ I just tried 0.4.10 and it's works, thanks !

Fifourche:
Oh and for the failing test, only on mac, I don't have a clue what might have triggered this error... Is this is something you already encountered @alisterburt ?

```
TEARDOWN ERROR: Exceptions caught in Qt event loop:
  ________________________________________________________________________________
  Traceback (most recent call last):
    File "/Users/runner/work/napari-animation/napari-animation/.tox/py39-macos-pyqt/lib/python3.9/site-packages/napari/_qt/utils.py", line 304, in remove_flash_animation
      widget.setGraphicsEffect(None)
  RuntimeError: wrapped C/C++ object of type QtWidgetOverlay has been deleted
  ```

**Comment by alisterburt** [comment]
oh that is looking nice!!

For the failing test, I've seen these errors before... usually they occur if I try to modify an attribute of a layer after closing the viewer... I'll relaunch the tests here to see if it passes on a second go!

**Comment by Fifourche** [comment]
> Was still on 0.4.8, didn't realize... ^^ I just tried 0.4.10 and it's works, thanks !

retried on the same computer and an other one, the problem is still there it seems... I'm not sure why ! I'll give a closer look at the `superqt` `QSliderLabel`and I might get back to you @tlambert03 !

Here are some informations about the packages' versions if needed :

```
napari: 0.4.10
Platform: Windows-10-10.0.19043-SP0
Python: 3.8.10 | packaged by conda-forge | (default, May 11 2021, 06:25:23) [MSC v.1916 64 bit (AMD64)]
Qt: 5.12.9
PyQt5: 5.12.3
NumPy: 1.21.1
SciPy: 1.7.0
Dask: 2021.07.0
VisPy: 0.7.1

OpenGL:
- GL version: 4.6.0 NVIDIA 462.31
- MAX_TEXTURE_SIZE: 32768

Screens:
- screen 1: resolution 1920x1080, scale 1.0

Plugins:
- animation: 0.0.3.dev89+g9d7b210
- console: 0.0.3
- scikit-image: 0.4.10
- svg: 0.1.5
```

tlambert03:
> retried on the same computer and an other one, the problem is still there it seems... 

when you previously [tried it](https://github.com/napari/napari-animation/pull/102#issuecomment-882586369) and thought it was working, was that a different OS?

tlambert03:
> a closer look at the superqt QSliderLabel

I actually don't think this is a superqt issue, it's more of a stylesheet issue, and the styles of widgets inserted into a napari viewer come from napari itself... superqt just "obeys" the parent styles.

The basic idea with layouts is that they get [allotted an area according to their size policy](https://doc.qt.io/qt-5/layout.html#adding-widgets-to-a-layout).  So, if they all have the default size policy, they will all get equal space (Which could give the appearance of a big gap like you had there).  The ways to fix it include:
- modifying the [size policy](https://doc.qt.io/qt-5/qsizepolicy.html#Policy-enum) of one or more widgets in the layout
- setting a fixed width either via the stylesheet or via `setFixedWidth`
- adding [stretch factors](https://doc.qt.io/qt-5/layout.html#stretch-factors)

Fifourche:
So from what I understood, the buttons of the `QDoubleSpinBox` (which is used to display the slider's value) are not displayed thanks to `setButtonSymbols(QSpinBox.NoButtons)`, but the buttons still occupy some space... I used a stylesheet to set their sizes to 0, and that seems to be ok ! 

Would there be some tests to write for this part of the interface then ? I see that codecov is failing, but at the same time there are no tests for rest of the plugin's interface either if I'm not mistaken !

tlambert03:
not totally following what does what there, but i pulled this and it looks good :)

One comment on saving:  currently, the entire napari window will block until saving is done.  This can go in another PR, but this would be a good time to use napari's `thread_worker` to put the saving into another thread, and instead of tqdm, you can use the [`progress`](https://github.com/napari/napari/blob/ac511362182a3bfbb43309fd216650269e6b10db/napari/_qt/qthreading.py#L662) parameter to have napari give you a progress bar in the activity window (rather than the console that launched napari, like tqdm).

nice work!

alisterburt:
apologies for taking so long about this @Fifourche, really let this get away from me! Congrats, awesome work 🚀

Issue #104, opened 2021-07-08
Changing the number of steps on the animation widget causes napari to crash if no frame has yet been captured


ptbrown1729:
Changing the number of steps on the animation widget causes napari to crash if no frame has yet been captured.

Minimal example: run the following code
```
import numpy as np
import napari
from napari_animation import AnimationWidget

viewer = napari.Viewer()

animation_widget = AnimationWidget(viewer)
viewer.window.add_dock_widget(animation_widget, area='right')
viewer.add_image(np.random.rand(301, 301, 301))
```
After napari launches, click the "+" button on the animation widget.
This gives the following error
```
WARNING: Traceback (most recent call last):
  File "c:\users\ptbrown2\documents\napari-animation\napari_animation\_qt\frame_widget.py", line 56, in _update_animation_steps
    active_keyframe.steps = self.stepsSpinBox.value()
AttributeError: 'NoneType' object has no attribute 'steps'
Process finished with exit code -1073740791 (0xC0000409)
```

napari: 0.4.10
Platform: Windows-10-10.0.17763-SP0
Python: 3.8.5 (default, Sep 3 2020, 21:29:08) [MSC v.1916 64 bit (AMD64)]
Qt: 5.9.7
PyQt5: 5.9.2
NumPy: 1.19.2
SciPy: 1.6.2
Dask: 2021.04.0
VisPy: 0.6.6

OpenGL:
- GL version: 4.6.0 NVIDIA 442.19
- MAX_TEXTURE_SIZE: 16384

Screens:
- screen 1: resolution 1920x1080, scale 1.0

Plugins:
- animation: 0.0.3.dev69+gaab5909
- console: 0.0.3
- scikit-image: 0.4.7
- svg: 0.1.4

alisterburt:
Hi @ptbrown1729 thanks for the clear report! 

Looks like a workaround for this would be to add 

```python
if active_keyframe is None:
    return
```
in the callback here between L55 and L56

https://github.com/napari/napari-animation/blob/aab59095ca0d4022014ff48b2c2a9620da20cdac/napari_animation/_qt/frame_widget.py#L53-L59

If you wanted to make this change yourself PRs are always welcome! Otherwise, thanks for taking the time to report and I'll make the change myself soon 🙂

ptbrown1729:
Sure, I'm happy to submit a PR

PR #105, opened 2021-07-08
fix for crash caused by changing number of frames

ptbrown1729:
Addressing issue #104, where changing the number of steps on the animation widget causes napari to crash if no frame has yet been captured.


alisterburt (approved):
Thanks for the PR, LGTM! 🚀

Issue #106, opened 2021-08-25
[BUG] Displayed data persists after visible is set to False


JoOkuma:
I'm trying to make a movie where some layers are turned on and off.

Making them visible is working, but sometimes it is not possible to make the layer invisible (it works in a few cases, but I couldn't narrow it down).

Minimal reproducing example:

```python3
import napari
from napari_animation import Animation

viewer = napari.Viewer()
viewer.open_sample('scikit-image', 'cells3d')

viewer.dims.ndisplay = 3
viewer.layers[0].visible = False

animation = Animation(viewer)
animation.capture_keyframe()

viewer.camera.angles = (0.0, 90, 180.0)
animation.capture_keyframe(steps=60)

viewer.layers[0].visible = True
viewer.camera.angles = (0.0, 180, 270.0)

animation.capture_keyframe(steps=60)

viewer.layers[0].visible = False
viewer.camera.angles = (0.0, 270, 0)

animation.capture_keyframe(steps=60)

animation.animate('demo.mov')
```

Video, I expected the membranes to be invisible after the second "bounce".

https://user-images.githubusercontent.com/21022743/130722072-67f6f598-9f31-4645-8c9f-dcc7ce9e28a5.mov

alisterburt:
Hi @JoOkuma - thanks for the nice and complete report, much appreciated! I'll take a look when I find some time but no guarantees of a timely fix. From twitter it looks like you managed to find a workable solution, do you have a better idea under what causes you were seeing this now? Is it possibly only happening at the end of an animation?

JoOkuma:
Hi, Alister. Sorry for the late response.

I found the bug this morning, the issue is with the [interpolation](https://github.com/napari/napari-animation/blob/07f93c7ae1fa2072110c822f5f791e4bd11a3296/napari_animation/interpolation.py#L27). 

I just learned that booleans are considered numbers in python so a boolean change of state calls the `interpolate_num` which results in a float between 0 and 1 that is interpreted as True except when it's zero, hence it's hardly ever updated.

I can change the if conditions to call `interpolate_bool` if one of them is `bool`, but this leads to the behavior of the change occurring in-between steps, for instance, in the above code the change happens after 30 steps

This is expected because it only changes when it crosses the 0.5 fraction, but shouldn't booleans changes happen instantly?
`

alisterburt:
@JoOkuma no worries! I had totally forgotten about this myself so thanks for coming back to it

[...]

PR #108, opened 2021-09-30
catch boolean interpolation earlier and change boolean interpolation behaviour

JoOkuma:
The problem was that booleans are numbers and when converted back to booleans the transition would almost never happen.

The code presented in issue #106, now results in the following movie:

https://user-images.githubusercontent.com/21022743/135507750-03433b42-5063-4311-af36-2a0b15024575.mov

An additional improvement was done, now booleans do not interpolate, their change is done instantaneously.

@alisterburt I updated one of the tests.
I'm not sure if the `interpolate_bool` is still an appropriate function name, I can change if you have any suggestion.

### DIFF
--- a/napari_animation/interpolation.py
+++ b/napari_animation/interpolation.py
@@ -9,7 +9,7 @@ from scipy.spatial.transform import Rotation as R
 
 def default(a, b, fraction):
     """Default interpolation for the corresponding type;
-    linear interpolation for numeric, step interpolation otherwise.
+    linear interpolation for numeric, instantaneous transition otherwise.
 
     Parameters
     ----------
@@ -24,13 +24,18 @@ def default(a, b, fraction):
     ----------
         Interpolated value between a and b at fraction.
     """
-    if isinstance(a, numbers.Number) and isinstance(b, numbers.Number):
+    if isinstance(a, bool) or isinstance(b, bool):
+        # checking this first because booleans are numbers
+        return interpolate_bool(a, b, fraction)
+
+    elif isinstance(a, numbers.Number) and isinstance(b, numbers.Number):
         return interpolate_num(a, b, fraction)
 
     elif isinstance(a, (list, tuple)) and isinstance(b, (list, tuple)):
         return interpolate_seq(a, b, fraction)
 
     else:
+        # strings, etc.
         return interpolate_bool(a, b, fraction)
 
 
@@ -74,7 +79,7 @@ def interpolate_num(a, b, fraction):
 
 
 def interpolate_bool(a, b, fraction):
-    """Step interpolation.
+    """Instantaneous transition from a to b.
 
     Parameters
     ----------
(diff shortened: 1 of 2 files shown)


JoOkuma:
It might be interesting to add an additional test similar to the code from the original issue, checking the state of the keyframes.

I can do that in the following weeks if you guys think it's a good idea.

napari_animation/interpolation.py
```diff
@@ -9,7 +9,7 @@
 
 def default(a, b, fraction):
     """Default interpolation for the corresponding type;
-    linear interpolation for numeric, step interpolation otherwise.
+    linear interpolation for numeric, instantaneous transition otherwise.
```
alisterburt, line 12:
```suggestion
    linear interpolation for numeric, step change at start of transition otherwise.
```

slightly more precise, feel free to modify if you feel strongly!

napari_animation/interpolation.py
```diff
@@ -24,13 +24,18 @@ def default(a, b, fraction):
     ----------
         Interpolated value between a and b at fraction.
     """
-    if isinstance(a, numbers.Number) and isinstance(b, numbers.Number):
+    if isinstance(a, bool) or isinstance(b, bool):
+        # checking this first because booleans are numbers
+        return interpolate_bool(a, b, fraction)
```
alisterburt, line 29: nice! very simple fix

napari_animation/interpolation.py
```diff
@@ -24,13 +24,18 @@ def default(a, b, fraction):
...
+
+    elif isinstance(a, numbers.Number) and isinstance(b, numbers.Number):
         return interpolate_num(a, b, fraction)
 
     elif isinstance(a, (list, tuple)) and isinstance(b, (list, tuple)):
         return interpolate_seq(a, b, fraction)
 
     else:
+        # strings, etc.
         return interpolate_bool(a, b, fraction)
```
alisterburt, line 39: do we know what actually gets caught in here? do we always want the instant transition at t0? I think we do but just trying to work out if this will lead to unexpected behaviour anywhere

napari_animation/interpolation.py
```diff
@@ -159,7 +164,10 @@ class Interpolation(Enum):
     SLERP = partial(slerp)
 
     def __call__(self, *args):
-        return self.value(*args)
+        # print(args)
+        val = self.value(*args)
+        # print(val)
+        return val
```
alisterburt, line 170:
```suggestion
        return self.value(*args)
```
looks like some debug prints got caught in here

alisterburt (approved):
Great work @JoOkuma - thanks for the fix!! Left a few small comments/questions, approving 🚀

alisterburt:
@JoOkuma re: the test! Yes that would definitely be useful to catch any regressions, in general this repo is not as well tested as it should be 🙂

napari_animation/interpolation.py
JoOkuma: Ops

napari_animation/interpolation.py
JoOkuma: Changes in opacity, translation, scale (? never tried it), are caught by the `interpolate_num` or the `seq` and later with `num`.

Changes in colormap and properties (? never tried neither) are caught in `else`. The issue is that they are not well defined with other interpolation operations. Changing them instantly I think covers most of the applications.

alisterburt: ideal - I think this is the expected behaviour then! :)

alisterburt:
not worried about the error here, seems internal napari only! merging now, thanks for the PR!!

PR #119, opened 2022-03-30
refactor viewer state interpolation machinery

alisterburt:
refactor and separation of everything to do with the interpolation of `ViewerState`

Added a fix for the error generating keys seen in #118 and allow us to move forwards there


napari_animation/_qt/frame_widget.py
```diff
@@ -4,7 +4,7 @@
 
 from qtpy.QtWidgets import QComboBox, QFormLayout, QSpinBox, QWidget
 
-from ..easing import Easing
+from napari_animation.easing import Easing
```
alisterburt, line 7:
```suggestion
from ..easing import Easing
```

Issue #122, opened 2022-08-21
Error when restarting napari-animation


haesleinhuepf:
When starting the napari animation wizzard twice (and closing the widget after the first attempt): The error below shows up. One can continue using napari-animation only after restarting napari. It does not matter if the plugin is called from the `Plugins` or the `Tools > Visualization` menu.

```
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
File C:\ProgramData\Miniconda3\envs\devbio-napari-env\lib\site-packages\napari\_qt\menus\plugins_menu.py:97, in PluginsMenu._add_plugin_actions.._add_toggle_widget(key=('animation', 'Wizard'), hook_type='dock')
     94     return
     96 if hook_type == 'dock':
---> 97     dock_widget, _w = self._win.add_plugin_dock_widget(*key)
        key = ('animation', 'Wizard')
        self._win = 
        self = 
     98 else:
     99     dock_widget = self._win._add_plugin_function_widget(*key)

File C:\ProgramData\Miniconda3\envs\devbio-napari-env\lib\site-packages\napari\_qt\qt_main_window.py:685, in Window.add_plugin_dock_widget(self=, plugin_name='animation', widget_name='Wizard')
    682         wdg = wdg._magic_widget
    683     return dock_widget, wdg
--> 685 wdg = _instantiate_dock_widget(Widget, self._qt_viewer.viewer)
        Widget = 
        self = 
    687 # Add dock widget
    688 dock_kwargs.pop('name', None)

File C:\ProgramData\Miniconda3\envs\devbio-napari-env\lib\site-packages\napari\_qt\qt_main_window.py:1248, in _instantiate_dock_widget(wdg_cls=, viewer=Viewer(axes=Axes(visible=False, labels=True, col....AnimationWidget object at 0x000002199561BEE0>>}))
   1243             break
   1244         # cannot look for param.kind == param.VAR_KEYWORD because
   1245         # QWidget allows **kwargs but errs on unknown keyword arguments
   1246 
   1247 # instantiate the widget
-> 1248 return wdg_cls(**kwargs)
[...]

PR #124, opened 2022-08-25
Overwrite existing keybindings during widget initialisation

cormsby626:
fix for #122


alisterburt (approved):
Thanks for the PR @cormsby626 - great work here!!

Issue #123, opened 2022-08-24
Grayscale and icons glitching after first recorded video frame


kephale:
I'm working on https://github.com/napari/napari/issues/4969, but am having some issues with the vanilla `napari-animation` on my M1. When recording one of the napari-animation examples (https://github.com/napari/napari/issues/4969#issuecomment-1224953981)...

The first frame records well:
<img width="1706" alt="image" src="https://user-images.githubusercontent.com/400105/186422507-933445d8-101f-4e5e-815f-c2e9ff35274f.png">

Then it becomes grayscale and some of the UI icons seem to go out of place:
<img width="1683" alt="image" src="https://user-images.githubusercontent.com/400105/186422801-5bbe48aa-00a4-4597-b262-db42b48a82c7.png">

@alisterburt does this happen on your M1 as well?

alisterburt:
hey @kephale ! Can you try pulling from main and seeing if this still happens? I think it might be the result of #111 not being fixed in the released version - I'm planning to spend a bit of time on this repo over the next couple evenings so will look into it more closely if it's still causing you trouble on main

alisterburt:
side note: is this normal for you? 
<img width="201" alt="image" src="https://user-images.githubusercontent.com/7307488/186436241-c82f1d59-aecb-43b0-9945-306efc164bf6.png">

kephale:
Hmm, I'm already working from main (commit a9555223bc7ca4cbeb5947d3d77e7ede056d6075).

alisterburt:
eep, will investigate then - I developed on my M1 but under rosetta, will test natively

kephale:
w.r.t. the question about the layout of the layer controls, the UI normally looks reasonable. Before recording that my napari was quite small, so maybe the layout didn't quite restore properly when running the script this time. I'm having a hard time recreating this interactively/manually though.

kephale:
It might be an issue with the `Labels` layer UI. If I add a `Points` layer that is active when I'm recording, then the UI looks normal.

<img width="939" alt="image" src="https://user-images.githubusercontent.com/400105/186498410-07cd25e1-01d7-4b75-881e-51e71d52a5d2.png">

alisterburt:
closed by #126

PR #126, opened 2022-08-30
Removed the flash during rendering

katherine-hutchings:
It looks like the removal of flashes added in #111 was lost at some point, I added it back in. 
@kephale I think this should fix your issue #123 , could you check please?


alisterburt (approved):
This is great @katherine-hutchings ! Thank you for taking the time 🚀

kephale:
image

@katherine-hutchings @alisterburt yep, this solves https://github.com/napari/napari-animation/issues/123. Thank you!

PR #144, opened 2022-12-16
Add max cache size to frame sequence

### LINKED ISSUE #143: napari-animation excessive memory usage
Hi, I had some problems running napari-animation with large datasets. It uses too much memory, sometimes even crashing due to running out-of-memory.

I think the problem is in [here](https://github.com/napari/napari-animation/blob/d8bd60939fb844e564ca54a5481cd0e2a80f854c/napari_animation/frame_sequence.py#L90-L111), where the `_cache` that keeps growing without limit.

JoOkuma:
Closes #143 

This is a quick solution, I did not dive deep into frame sequence logic. It might be better to implement using python [`lru_cache`](https://docs.python.org/3/library/functools.html#functools.lru_cache).

If that's the case I can update it in a few days when I have more time.

### DIFF
--- a/napari_animation/frame_sequence.py
+++ b/napari_animation/frame_sequence.py
@@ -34,7 +34,7 @@ class FrameSequence(Sequence[ViewerState]):
         A KeyFrameList from which to render the final frame sequence.
     """
 
-    def __init__(self, key_frames: KeyFrameList) -> None:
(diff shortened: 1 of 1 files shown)


alisterburt:
thanks for the PR @JoOkuma ! I don't love the implementation so will followup with an lru_cache appraoch over the break :)

Issue #147, opened 2023-03-01
Using Jupyter Output to display videos


kolibril13:
It would be nice to have a function or a cellmagic that displays the generated video directly in the Jupyter output!

Here is a prototype:

*Cell1*
```py
import napari
from skimage import data
import numpy as np
from IPython.display import Video

cat = data.cat()


viewer = napari.view_image(cat, rgb=True)


def jupyter_napari_animation(animation,**kwargs):
    from pathlib import Path
    from datetime import datetime

    dir = Path.cwd() / "media"
    dir.mkdir(exist_ok=True)
    mov_name =  "napari_video" + "@" + datetime.now().strftime("%Y-%m-%d@%H-%M-%S") + ".mp4"

    path = dir / mov_name
    animation.animate(path, **kwargs)
    #  ^ I know, the above line should be done differently! e.g. jupyter_napari_animation could be a method of the animation object.

    vid = Video(path, width=600, html_attributes=f"controls autoplay loop")
    display(vid)
```

*Cell2*
```python
from napari_animation import Animation

animation = Animation(viewer)
viewer.camera.zoom = 1
animation.capture_keyframe(steps=0)
viewer.camera.zoom = 2
animation.capture_keyframe(steps=30)

jupyter_napari_animation(animation, canvas_only=True, fps=30)
```

This is inspired by the way manim videos are embedded:

https://github.com/ManimCommunity/manim/blob/3713ed6b1e0f9d51de8e74717bf4489c592f880f/manim/utils/ipython_magic.py#L176-L197

alisterburt:
agree this would be cool! It looks like animate doesn't currently return anything - we could have the method return a small object which implements [_ipython_display()](https://ipython.readthedocs.io/en/stable/config/integrating.html#more-powerful-methods) and shows the video as you suggest! :)

I don't need this myself or use jupyter notebooks particularly heavily but would gladly accept a PR if you're keen to make this happen!

kolibril13:
Glad you like it! :)

> It looks like animate doesn't currently return anything - we could have the method return a small object which implements [_ipython_display()](https://ipython.readthedocs.io/en/stable/config/integrating.html#more-powerful-methods) and shows the video as you suggest! :)

In theory, a good idea! In practive, I remember one issue that we had in manim is, that when you name the video yourself, e.g. animation.animate("my_vid.mov"), then the video did not update due to caching issues.
[...]

PR #148, opened 2023-03-01
Add ease func example in notebook

kolibril13:
This pr adds a notebook with an ease func.
Note that the video output is embedded, so the notebook is quite big (0.3 MB in total)


alisterburt (commented):
Thanks @kolibril13 - could you add this example as .py instead? not a huge fan of notebooks in repos :)

kolibril13:
Yes, I can see your point and I will convert it to .py! :)

Just to defend notebooks a bit:
I think they are absolutely great, even in repos.
And there is even a new GitHub Feature coming soon that will allow rich diff:
image

Also, I think that napari-animation would be amazing to be used in notebooks, as mentioned here: https://github.com/napari/napari-animation/issues/147

kolibril13:
file is not converted.

alisterburt (approved):
wow, that feature looks awesome!! Thanks for the quick turnaround here

PR #150, opened 2023-03-02
Limit cache size in frame sequence

jni:
Alternative to #144 

I've implemented a very minimal LRUDict to replace the cache.


alisterburt (approved):
Yer a wizard @jni - looks great and thanks for taking the time!

alisterburt:
I need to sort CI but will optimistically merge here and do that separately!

PR #153, opened 2023-03-02
add example notebook

kolibril13:
This notebook is an example of how napari-animation can be used in notebooks.
A nice benefit I can see from using notebooks is that the viewer is only initialized once in the beginning, and can be recycled for all following animations.

Related to https://github.com/napari/napari-animation/issues/147


alisterburt:
Cool! Could you add the word 'programmatic' to the filename before merge? :)

Cheers!

kolibril13:
Done!

alisterburt:
thanks! merging 🚀

PR #160, opened 2023-03-19
fix tox

kevinyamauchi:
This PR updates the `tox.ini` so that the passenv items are all on individual lines.


alisterburt:
thanks for this @kevinyamauchi !

PR #167, opened 2023-04-20
Include napari version information in output video metadata

GenevieveBuckley:
## Description

Companion to PR https://github.com/napari/napari/pull/5494

This adds a digital watermark to the video files output by napari-animation. That means in the future it may be possible to scrape publications & supplementary data, eg: like this analysis done for matplotlib https://github.com/mrocklin/arxiv-matplotlib

## Try it out!

Now, you can run the examples to create a video file:
```python
python napari-animation/examples/animate2d.py
```
...and then look at the metadata embedded in the video file header with ffmpeg. This ffmpeg command writes the metadata to a text file:
```
!ffmpeg -i demo2d.mov -f ffmetadata FFMETADATAFILE.txt
```
And when you look at the output
```
cat FFMETADATAFILE.txt
```
...you see something like this (note the new "title" metadata field, which was not there before:
```
;FFMETADATA1
major_brand=qt  
minor_version=512
compatible_brands=qt  
title="napari version 0.4.17 https://napari.org/"
encoder=Lavf59.27.100
```

Actually, here is a simpler way to read the metadata from the video file:
```python
    # Read metadata back in, and check for napari version information
    # We expect to see a metadata line in the metadata like this:
    # title="napari version 0.4.17 https://napari.org/"
    with open(output_video_filename, 'rb') as f:
        content = f.read()
    assert b"napari version" in content
    assert b"https://napari.org" in content
```

## Additional details

I have stuffed the napari version information into the "title" metadata key, because that seems to be the most widely available. (.mp4 files do allow a "description" metadata field, but that is *not* supported by .avi, .mov, and possibly many other file formats).

### DIFF
--- a/napari_animation/_tests/test_animation.py
+++ b/napari_animation/_tests/test_animation.py
@@ -1,5 +1,3 @@
-import os
-import subprocess
 from unittest.mock import patch
 
 import numpy as np
@@ -119,13 +117,12 @@ def test_animation_file_metadata(animation_with_key_frames, tmp_path, ext):
     output_filename = tmp_path / f"test{ext}"
     animation.animate(output_filename)
     # Read metadata back in, and check for napari version information
-    output_metadata_filename = os.path.splitext(output_filename)[0] + ext.replace('.', '-') + ".txt"
-    subprocess.run(f"ffmpeg -i {output_filename} -f ffmetadata {output_metadata_filename}", shell=True, check=True)
-    with open(output_metadata_filename) as f:
-        content = f.read()
     # We expect to see a metadata line in the metadata like this:
     # title="napari version 0.4.17 https://napari.org/"
-    assert 'title="napari version' in content
+    with open(output_filename, 'rb') as f:
+        content = f.read()
+    assert b"napari version" in content
+    assert b"https://napari.org" in content
 
 
 @pytest.mark.parametrize("attribute", CAPTURED_LAYER_ATTRIBUTES)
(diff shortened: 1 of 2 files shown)


napari_animation/_tests/test_animation.py
```diff
@@ -110,6 +112,22 @@ def test_animate_filenames(
         assert saved_files == expected
 
 
+@pytest.mark.parametrize("ext", [".mp4", ".mov", ".avi", ""])
+def test_animation_file_metadata(animation_with_key_frames, tmp_path, ext):
+    """Test output video file contians napari version metadata()"""
+    animation = animation_with_key_frames
+    output_filename = tmp_path / f"test{ext}"
+    animation.animate(output_filename)
```
GenevieveBuckley, line 120: I notice that the other tests (like `test_animate_filenames` above) are very careful to mock/patch the video writer, so I'm not sure if you won't like that it actually runs here.
I'm not really sure how best to test this without generating real output video files though.

alisterburt: I think this is okay, that was done because the output wasn't being checked for anything afaik

napari_animation/animation.py
```diff
@@ -5,6 +5,7 @@
 
 import imageio
 import numpy as np
+from napari._version import __version__
```
alisterburt, line 8:
```suggestion
from napari._version import __version__ as napari_version
```

napari_animation/animation.py
```diff
@@ -155,6 +156,9 @@ def animate(
         """
         self._validate_animation()
 
+        description = f"napari version {__version__} https://napari.org/"
```
alisterburt, line 159:
```suggestion
        description = f"napari version {napari_version} https://napari.org/"
```

alisterburt (approved):
Cool! thanks for the PR - made a tiny change so there wasnt a spurious reference to `__version__` within the package but otherwise this is great!

GenevieveBuckley:
Thanks for the review!

PR #170, opened 2023-05-01
Easily display video output in jupyter notebooks

GenevieveBuckley:
Closes https://github.com/napari/napari-animation/issues/147

This PR adds a `_repr_html_` method to the `Animation` class, so that it will automatically display the video once the filename is known (i.e. after the `animate` method has been run).

Screen Shot 2023-05-01 at 5 11 16 pm

### DIFF
--- a/napari_animation/animation.py
+++ b/napari_animation/animation.py
@@ -248,5 +248,8 @@ class Animation:
             self.set_key_frame_index(keyframe_index)
 
     def _repr_html_(self):
-        html = f'<video width="100%" height="100%" controls> <source src="{self._filename}"> </video>'
-        return html
+        if self._filename is None:
+            return None
+        else:
+            html = f'<video width="100%" height="100%" controls> <source src="{self._filename}"> </video>'
+            return html
--- a/napari_animation/animation.py
+++ b/napari_animation/animation.py
@@ -249,7 +249,7 @@ class Animation:
 
     def _repr_html_(self):
         if self._filename is None:
-            return None
+            return 'Video animation not yet available (use the "animate" method to generate it).'
         else:
             html = f'<video width="100%" height="100%" controls> <source src="{self._filename}"> </video>'
             return html
--- a/napari_animation/animation.py
+++ b/napari_animation/animation.py
@@ -45,6 +45,8 @@ class Animation:
 
         self._frames = FrameSequence(self.key_frames)
 
+        self._filename = None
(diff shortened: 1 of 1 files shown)


GenevieveBuckley:
Note: I did try Alister's [ipython_display suggestion](https://github.com/napari/napari-animation/issues/147#issuecomment-1450898803), as well as `_repr_html_`, but that didn't actually work very well for me. 

Perhaps this was because I was running jupyter lab instead of a jupyter notebook, but in any case I decided to go with the one that seemed more robust.

napari_animation/animation.py
```diff
@@ -243,3 +246,10 @@ def _on_active_keyframe_changed(self, event):
         if active_keyframe:
             keyframe_index = self.key_frames.index(active_keyframe)
             self.set_key_frame_index(keyframe_index)
+
+    def _repr_html_(self):
+        if self._filename is None:
+            return 'Video animation not yet available (use the "animate" method to generate it).'
+        else:
+            html = f'<video width="100%" height="100%" controls> <source src="{self._filename}"> </video>'
```
GenevieveBuckley, line 254: I know it seems silly to have both height and width set to 100%, but this actually seems to generate nice responsive and resize-able video windows as output, without me having to guess whether the video is likely to be portait or landscape (and also whether the notebook screen space is portrait or landscape).

napari_animation/animation.py
```diff
@@ -243,3 +246,10 @@ def _on_active_keyframe_changed(self, event):
         if active_keyframe:
             keyframe_index = self.key_frames.index(active_keyframe)
             self.set_key_frame_index(keyframe_index)
+
+    def _repr_html_(self):
+        if self._filename is None:
+            return 'Video animation not yet available (use the "animate" method to generate it).'
```
GenevieveBuckley, line 252: This is the message you see if the animation has not yet been saved. I think it's clear, but welcome suggestions to improve it.

Screen Shot 2023-05-01 at 5 39 21 pm

alisterburt:
awesome - thanks for the PR, this looks really great! My only question is whether this needs to be adjusted if an animation is saved as a .gif? Happy for this to be handled in followup though - will merge here :)

alisterburt (approved):

alisterburt:
@GenevieveBuckley FYI I just pushed v0.0.6 which includes this :)

PR #182, opened 2023-08-17
change fps to duration to account for changes to imageio

aelefebv:


aelefebv:
Alternatively, we could keep fps as a parameter and translate it to duration before passing it to imageio via duration=1000/fps

jni:
> Alternatively, we could keep fps as a parameter and translate it to duration before passing it to imageio via duration=1000/fps

Personally this would be my preference, especially because you could gate that on the imageio version, so there would be no disruption at all to any users (even those with older imageio).

jni:
I also think it's easier to think about fps

alisterburt:
Thanks for the PR @aelefebv ! Much appreciated :-) 

I agree with Juan here, we should not mirror the imageio API and should instead keep ours consistent with FPS

aelefebv:
Makes sense to me! Just changed it back to fps and now passing in duration to get_writer.

And thanks again for the great plugin :)

jni:
Hmmm, perhaps we need a try/except or if/else somewhere to account for multiple imageio versions? The tests are failing with the current changes:

```
   E       TypeError: FfmpegFormat.Writer._open() got an unexpected keyword argument 'duration'
```

napari_animation/animation.py
```diff
@@ -187,15 +188,15 @@ def animate(
                 ]:
                     writer = imageio.get_writer(
                         filename,
-                        fps=fps,
+                        duration=duration,
```
psobolewskiPhD, line 191:
```suggestion
                        fps=fps,
```
Only the pillow plugin is using duration

psobolewskiPhD:
@jni The `duration` kwarg is for pillow plugin only (see https://github.com/imageio/imageio/issues/992#issuecomment-1567479878 ) so I just reverted the change for ffmpeg. I think this is good to go, unless you prefer to not have the duration setting in the `try:` -- could just do it in the `imageio` call.

psobolewskiPhD (approved):

Issue #168, opened 2023-04-20
How to run the tests locally?


GenevieveBuckley:
I'm having trouble testing this package locally.
* With pytest, it complains that qtbot is not available for the `make_napari_viewer` fixture. I tried installing qtbot and giving things another go, but that didn't help.
* With tox
    * tox on Windows - it finds the one matching tox configuration that matches the local environment, but hangs for a really long time when installing the dependencies. I've left it going, but I don't think it's going to actually run. *EDIT: I left it overnight and it did finally run, hooray! Nevertheless, I usually work on a Mac, so this might not help me that much.*
    * tox on Mac - all tox configurations are skipped, and I get these errors
```
py310-macos-pyqt: failed with pass_env values cannot contain whitespace, use comma to have multiple values in a single line, invalid values found 'DISPLAY XAUTHORITY'
py310-macos-pyqt: FAIL ✖ in 0 seconds
py310-macos-pyside: failed with pass_env values cannot contain whitespace, use comma to have multiple values in a single line, invalid values found 'DISPLAY XAUTHORITY'
```

The README says *"...a detailed contributing guide is coming soon."*, so perhaps whatever the answer to this problem is could become part of the not-yet-existing developer contributing guide?

alisterburt:
ooh - sorry about this

I'm also on a mac and running tests with pytest
```
platform darwin -- Python 3.10.6, pytest-7.1.3, pluggy-1.0.0
PyQt5 5.15.7 -- Qt runtime 5.15.4 -- Qt compiled 5.15.4
rootdir: /Users/alisterburt/programming/napari-animation, configfile: setup.cfg
plugins: qt-4.1.0, napari-0.4.16, napari-plugin-engine-0.2.0, npe2-0.6.1
collected 97 items                            
```

What is the error you're getting? How are you installing `pytest-qt`?

alisterburt:
agree that this would be great in a contributing guide and sorry for the false advertising in the readme - is leftover from when we were pushing hard on this and had grand plans 😆

GenevieveBuckley:
> I'm also on a mac and running tests with pytest

Can I see your (working) conda list? That might be helpful

GenevieveBuckley:
First I tried this...
```
conda create -n napari-animation-dev python=3.10 pip ipython
conda activate napari-animation-dev
python -m pip install -e ".[dev]"
```

But that gave me an error installing pyqt.
<details>
<summary>Details (click to expand)</summary>

```
[...]

PR #188, opened 2023-11-25
[Maint] Improve contribution in README and remove pyqt5 from [dev] install dependency

psobolewskiPhD:
Closes: https://github.com/napari/napari-animation/issues/185
Closes: https://github.com/napari/napari-animation/issues/168

I've removed pyqt5 from [dev] and added a Note in the Contributing section that you need a suitable Qt backend for napari.  I've added an example using a conda virtual env and installing `[dev]` and PyQt5. Finally, I explicitly provide the command to run tests locally.

### DIFF
--- a/README.md
+++ b/README.md
@@ -88,23 +88,33 @@ use of the plugin follow.
 ## Contributing
 
 Contributions are very welcome and a detailed contributing guide is coming soon.
-In the meantime, clone this repository and install it in editable mode:
-`pip install -e .[dev]`
-> [!NOTE]
-> Ensure you have a suitable Qt backend for napari!
-
-Tests are run with `pytest`.
-
-We use [`pre-commit`](https://pre-commit.com) to sort imports with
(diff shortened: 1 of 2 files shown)


jni (approved):

jni:
@psobolewskiPhD should we link to https://napari.org/dev/tutorials/fundamentals/installation.html?highlight=qt%20backend#choosing-a-different-qt-backend?

I'm using the dev link because backends seem to be something where updates can happen independently of the napari version.

psobolewskiPhD:
Sure I can add that. The gist is you need a functioning napari install, which means having a Qt backend. But if you have napari with pyside or pyqt6 or whatever this plugin shouldn't override that with pyqt5.

psobolewskiPhD:
I updated the PR (and OP) to add the requested link and to address #168 also.
The shell block followed by `pytest .` works for me locally--all green!--with this PR.

jni (approved):

PR #195, opened 2024-01-10
Use boolean interpolation for `dims.ndisplay`

psobolewskiPhD:
Closes: https://github.com/napari/napari-animation/issues/156

This PR adds `interpolate_bool` to the Interpolation ENUM and then uses that for interpolating `dims.ndisplay` because this can only be 2 or 3, behaving like a boolean, which is how all strings are interpolated.
Previously, the numeric interpolation was used because `ndisplay` type is int and so the outcome was cast to int, which truncated the decimal resulting in the opposite behavior from strings, making for a mismatch in the end behavior.

Try doing an animation where you change to 3D viewer and then rotate the object. On live, this won't result in the expected animation, but with this PR it will.

```
import napari
from napari_animation import Animation

viewer = napari.Viewer()
viewer.dims.ndisplay = 2
viewer.open_sample("napari", "binary_blobs_3D")
anim = Animation(viewer)
anim.capture_keyframe(steps=3)
viewer.dims.ndisplay = 3
viewer.camera.angles = (-30, 40, 15)
anim.capture_keyframe()
anim.animate("test.mp4")
```

### DIFF
--- a/napari_animation/_tests/conftest.py
+++ b/napari_animation/_tests/conftest.py
@@ -42,7 +42,7 @@ def frame_sequence(animation_with_key_frames):
 @pytest.fixture
 def image_animation(make_napari_viewer):
     viewer = make_napari_viewer()
-    viewer.add_image(np.random.random((28, 28)))
+    viewer.add_image(np.random.random((2, 28, 28)))
     return Animation(viewer)
 
 
--- a/napari_animation/_tests/test_animation.py
+++ b/napari_animation/_tests/test_animation.py
@@ -151,6 +151,16 @@ def test_end_state_reached(image_animation):
     assert last_state == image_animation.key_frames[-1].viewer_state
 
 
+def test_ndisplay_interpolation(image_animation):
+    """Check that dims.ndisplay interpolation is boolean"""
+    image_animation.capture_keyframe()
+    image_animation.viewer.dims.ndisplay = 3
+    image_animation.capture_keyframe(steps=3)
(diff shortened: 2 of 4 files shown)


jni (approved):

jni:
Ah, I guess in my example I bookended the ndisplay change with frame captures, so interpolation didn't matter. My spidey sense was "I don't know what is going to be done here so imma make sure I capture around it." 😂 

I'm a little confused about the syntax of the script. Specifically the lines:

```python
anim.capture_keyframe(steps=3)
viewer.dims.ndisplay = 3
viewer.camera.angles = (-30, 40, 15)
anim.capture_keyframe()
```

But if I look at the `capture_keyframe` docstring, the documentation for steps is:

> steps : int
> Number of interpolation steps between last keyframe and captured one.

So as far as I understand it, there should only be a single 3D frame, since the capture with 3 frames happens before the viewer is 3D? Am I misunderstanding something or is this exposing another bug / a documentation error?

psobolewskiPhD:
I'm not sure I get the question?
The first keyframe is set with no change of the viewer, so I used `3` to save on file size.
The next keyframe, uses the default 15, but yes before this PR yes, you will get 1 frame of 3D, no matter what steps.
With this PR you see the switch to 3D in 1 frame and then 14 frames of the rotation.

jni:
Yes, I'm an idiot, I didn't realise the default was 15 and was reading it as the default being 1, hence my confusion. 🤦 Thanks for your patience with me @psobolewskiPhD! 😅

PR #197, opened 2024-01-12
[DOC] Warn about `ffmpeg` on arm64 macOS and add Conda to installation instructions

psobolewskiPhD:
Closes: https://github.com/napari/napari-animation/issues/196

The PyPI package doesn't have an arm64 macOS binary for ffmpeg, so exporting animations doesn't work with a pip install. I've added a warning about this to the README and provide two ways to get arm64 ffmpeg installed which both work.
Additionally, add section for Conda installation.

### DIFF
--- a/README.md
+++ b/README.md
@@ -39,8 +39,8 @@ pip install napari-animation
 ```
 
 > [!WARNING]
-> `napari-animation` uses `ffmpeg` to export animations. If you are using a macOS arm64 computer (e.g. M1, M2 processor)
-> the PyPI package does not include the needed binary for your platform. You will need to install `ffmpeg` from using
+> `napari-animation` uses `ffmpeg` to export animations. If you are using a macOS arm64 computer (Apple Silicon e.g. M1, M2 processor)
+> the PyPI package does not include the needed binary for your platform. You will need to install `ffmpeg` using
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -38,6 +38,19 @@ This plugin remains under development and contributions are very welcome, please
 pip install napari-animation
 ```
 
+> [!WARNING]
+> `napari-animation` uses `ffmpeg` to export animations. If you are using a macOS arm64 computer (e.g. M1, M2 processor)
```
GenevieveBuckley, line 42: You might like to also include "Apple Silicon" as a keyword here. Some users may not recognize synonyms, so it's helpful to include.

README.md
```diff
@@ -38,6 +38,19 @@ This plugin remains under development and contributions are very welcome, please
...
+> the PyPI package does not include the needed binary for your platform. You will need to install `ffmpeg` from using
```
GenevieveBuckley, line 43:
```suggestion
> the PyPI package does not include the needed binary for your platform. You will need to install `ffmpeg` using
```

Delete either "from" or "using" at the end of this sentence. Doesn't matter which one, just can't have both.

GenevieveBuckley: ...but since I see we have "from" later on in that sentence, it makes sense to say "using conda, from the conda-forge channel..."

GenevieveBuckley (approved):

PR #198, opened 2024-01-12
[MAINT] Add python 3.11 to test matrix

psobolewskiPhD:
Adding python 3.11 to test matrix, which has experimental support in napari.
Once napari drops python 3.8 (real soon now? see https://github.com/napari/napari/pull/5768 ) then we can also drop it here.
Note that pyside2 doesn't have any current wheels for python 3.11, so had to split pyqt and pyside testing to not run pyside2 with python 3.11
Also had to bump the black version to support python 3.11


GenevieveBuckley:
Looks great, thanks @psobolewskiPhD

PR #199, opened 2024-01-12
[MAINT] switch to ruff, per napari/napari

psobolewskiPhD:
Just to make life easier, switching this repo to also use ruff like napari/napari, instead of flake8 and isort.
Also dropped some old Conda cruft from tox.ini which wasn't used.

After making the tooling changes, I also ran ruff and made various fixes to get everything to pass.


tox.ini
```diff
@@ -26,11 +26,6 @@ passenv =
     XAUTHORITY
     NUMPY_EXPERIMENTAL_ARRAY_FUNCTION
     PYVISTA_OFF_SCREEN
-conda_deps =
-    # use conda to install numcodecs on mac py3.9
-    py39-macos: numcodecs
-conda_channels =
```
GenevieveBuckley, line 32: Why do conda_deps and conda_channels need to be removed here? I don't quite understand how that is related to ruff.

GenevieveBuckley (approved):
This looks good to me. 
Since the ruff CI test is passing, I say go ahead and merge.

tox.ini
Czaki: I thinkt hat it comes from this part of description:

>  Also dropped some old Conda cruft from tox.ini which wasn't used.

GenevieveBuckley: No need for it to be a separate PR, just checking that you intended this 😄

GenevieveBuckley:
Great, thank you for this @psobolewskiPhD 🎉

PR #200, opened 2024-01-25
Documentation set up with Sphinx

melissawm:
Supersedes #173 

May close #80 

I haven't really looked at the scrapers, because currently the docs execute even if the images are not in the gallery. Feel free to push to this PR if you know how to build the animation scraper. I can also try to come up with a solution later.

cc @GenevieveBuckley


docs/developers/contributing.md
```diff
@@ -0,0 +1,171 @@
...
+```sh
+make docs
+```
+
+The docs will be built at `docs/_build/html`.
+
+Most web browsers will allow you to preview HTML pages.
+Try entering `file:///absolute/path/to/napari/docs/_build/html/index.html` in your address bar.
+
+To read more about the docs, how they're organized, and built, read {ref}`docs-dev`.
```
GenevieveBuckley, line 160: When I build the documentation it says this reference {ref}`docs-dev` is undefined. What do we need to do to fix that?

docs/developers/contributing.md
```diff
@@ -0,0 +1,171 @@
...
+
+## Code of conduct
+
+`napari` has a [Code of Conduct](CODE_OF_CONDUCT.md) that should be honored by everyone who participates in the `napari` community.
```
GenevieveBuckley, line 164: The napari-animation repository does not include a code of conduct file. Should we just make a copy of the `CODE_OF_CONDUCT.md` from the main napari repository to get rid of this link error, or do you prefer another solution?

docs/index.md
```diff
@@ -0,0 +1,135 @@
+# napari-animation documentation
```
GenevieveBuckley, line 1: I really don't like that this `docs/index.md` file is an almost-identical copy of the `README.md`. This is a recipe for updates/changes being made in one place but not the other, and eventually ending up with users/developers having only partial information (depending on whether they mostly look at the built docs website, or mostly look at the github repo home page).

Can we change this, either to point sphinx directly to the main README.md in the repository, or somehow have the contents actually copied across *during* the build process?

setup.cfg
```diff
@@ -50,6 +50,17 @@ dev =
...
     check-manifest
     %(testing)s
+doc =
+    sphinx>6
+    sphinx-autobuild
+    sphinx-external-toc
+    sphinx-copybutton
+    sphinx-gallery
+    sphinx-favicon
+    sphinxcontrib-video
```
GenevieveBuckley, line 60: FYI I am using the sphinxcontrib-video plugin for convenience, although we don't have to if we prefer not. https://github.com/sphinx-contrib/video

docs/conf.py
```diff
@@ -0,0 +1,336 @@
...
+# linkcheck_anchors_ignore = [r'^!', r'L\d+-L\d+', r'r\d+', r'issuecomment-\d+']
+
+# linkcheck_ignore = [
+#     'https://napari.zulipchat.com/',
+#     '../_tags',
+#     'https://en.wikipedia.org/wiki/Napari#/media/File:Tabuaeran_Kiribati.jpg',
+# ]
+
+
+# def rewrite_github_anchor(app, uri: str):
```
GenevieveBuckley, line 313: I don't love that there are big chunks of commented code in this file (linkcheck_ignore, rewrite_github_anchor, etc.). I know you just copy-pasted the file contents from the napari/docs repo, but what's happening here? Do we 
1. need to remove these lines entirely,
2. or does there need to be an issue opened about what problem needs to be fixed here?

docs/conf.py
```diff
@@ -0,0 +1,336 @@
...
+                video_names.append(this_video_path)
+                shutil.move(video, this_video_path)
+                self.video_thumbnail(this_video_path)
+                rst += self.rst_video_template(this_video_path)
+
+        # Use the `figure_rst` helper function to generate reST for image files
+        # result = figure_rst(video_names, gallery_conf['src_dir'])
+        return rst
+
+    def video_thumbnail(self, video_filename):
```
GenevieveBuckley, line 211: I think we need this so we end up with the little static thumbnail pictures in the example gallery overview page.

docs/conf.py
```diff
@@ -0,0 +1,336 @@
...
+        return "VideoScraper"
+
+    def __call__(self, block, block_vars, gallery_conf):
+        """Video file scraper.
+        Scrapes video files that were saved to disk by the example scripts.
+        Based on the sphinx example scraper "Example 2: detecting image files on disk"
+        https://sphinx-gallery.github.io/stable/advanced.html#example-2-detecting-image-files-on-disk
+        """
+        # Find all video files in the directory of this example.
+        video_file_extensions = [".mp4", ".mov"]
```
GenevieveBuckley, line 182: I checked this works for .mp4 and .mov filetypes. We could add more, so long as someone checks they work well in the built docs.

docs/conf.py
```diff
@@ -0,0 +1,336 @@
...
+        first_frame_filename = os.path.splitext(video_filename)[0] + ".png"
+        iio.imwrite(first_frame_filename, first_frame)
+        return first_frame
+
+    def rst_video_template(self, video_filepath):
+        """Template HTML for embedding video into webpage."""
+        rst = f""".. video:: {video_filepath}
+        :autoplay:
+        :loop:
+        :width: 600
```
GenevieveBuckley, line 224: I don't like this part. Currently it sets the width of the embedded video to 500 pixels. I'd prefer it was a percentage of the page width instead (i.e. more responsive to different screen sizes). I don't think you can do that easily with sphinxcontrib-video as it currently exists, so maybe we should replace this dependency with a template docstring of the HTML we need to embed in the RST document, so we can have more control over the output result.
Docs: https://github.com/sphinx-contrib/video

GenevieveBuckley:
I don't see the ipython notebook example int he built html docs. I had expected to see it appear as a page in the examples gallery. @melissawm is this part working for you?

docs/developers/contributing.md
GenevieveBuckley: Update: I've linked to the canonical webpage at https://napari.org/stable/community/code_of_conduct.html

docs/conf.py
GenevieveBuckley: Although for some reason, when I build the documentation with `make docs` there are warnings output about the .mov file format. 

I'm not sure why, because the HTML website works as expected, regardless of whether we have .mov or .mp4 videos embedded. Regardless, I've changed the output files from our example scripts to use .mp4, just to avoid the warnings in the docs build.

docs/conf.py
```diff
@@ -0,0 +1,336 @@
...
+                rst += self.rst_video_template(this_video_path)
+
+        # Use the `figure_rst` helper function to generate reST for image files
+        # result = figure_rst(video_names, gallery_conf['src_dir'])
+        return rst
+
+    def video_thumbnail(self, video_filename):
+        """Save PNG screenshot of the first video frame."""
+        reader = iio.get_reader(video_filename)
+        first_frame = reader.get_data(0)
```
GenevieveBuckley, line 214: Note that the image is from the first frame of the video.

That means in some cases it isn't a very interesting picture. Like the layer planes example, which fades in from black, so you don't see anything very exciting in the thumbnail image.

I tried to make this function choose a frame from the middle of the videos, but ffmpeg is returning "Inf" lengths for these movie files, so I have no good way to know where the halfway point is.

docs/conf.py
melissawm: Sorry I wasn't clear - these are the scrapers from napari, and I left them here for reference. I don't know if they would work for animations, and I would need help to figure that out. Since you said you had done this before, I was hoping for a suggestion or two 😅

docs/developers/contributing.md
melissawm: I got this file from Alister's previous PR and didn't review properly. Will fix.

docs/index.md
melissawm: Sure, I didn't know what to write here so just copied the README as a placeholder. I can do an include, but there are some sections that won't work, for example the admonitions (since github markdown and myst markdown don't agree on that syntax)

GenevieveBuckley:
Summarizing the current status (Melissa, please add or correct anything I've got wrong here)

Items remaining for this PR:
- [ ] Add Github actions script to build and deploy docs (presumably will be based on [this docs deploy script](https://github.com/napari/docs/blob/main/.github/workflows/deploy_docs.yml))
- [ ] Set up Github secret key for docs deployment (apparently I don't have the right permissions for this - @jni who does?)
- [ ] Resolve README duplication ([discussion thread here](https://github.com/napari/napari-animation/pull/200#discussion_r1465841228))
- [ ] Are the `linkcheck_ignore` and `rewrite_github_anchor` functions staying or going ([see discussion thread here](https://github.com/napari/napari-animation/pull/200#discussion_r1465843384))


Nice to have, possibly in a later PR:
- [ ] I can modify the VideoScraper class to close any open napari windows *if* an output video file is found, then run the regular static napari screenshot scraper for all other cases. Is that behaviour we'd want? (It's easy for me to do, I just want to know if that's what other people think should happen)
- [ ] Responsive (or percentage based) sizing of embedded videos. This currently isn't supported by the sphinx video plugin, so we'd have to either modify that or write our own HTML template string. ([See here](https://github.com/napari/napari-animation/pull/200#discussion_r1465848115))
- [ ] Executing .ipynb jupyter notebook files. I'm not sure what the reason for skipping these in the napari/docs is. There might be some tricky technical issue preventing it here too? See [this line of code in conf.py](https://github.com/napari/docs/blob/d6fa7d07f0e72325fff6e3d3fa5cd66f3790bbd9/docs/conf.py#L294).

jni:
>  Set up Github secret key for docs deployment (apparently I don't have the right permissions for this - @jni who does?)

I don't think you actually need this to deploy to the same-repo's gh-pages. See the [deploy to github pages docs](https://github.com/marketplace/actions/deploy-to-github-pages) (though they are not as clear as I'd like about this):

> Alternatively you can set the following in your workflow file to grant the action the permissions it needs.
>
> ```yaml
> permissions:
>   contents: write
> ```

To answer your Q, the napari steering council can do that if needed. But I think it's not. 😅

docs/index.md
jni: >  1B) Have bad admonition formatting on the github page. Personally, I'd prioritize making the official website look good.

This would be my choice also.

GenevieveBuckley:
Ok, maybe something is funky about how the napari/docs repository handles this, that's where the secret key business came from. If someone would like to edit the new `deploy_docs.yml` file here to remove that requirement, that would be fantastic.

jni:
Yeah the issue with the napari/docs repo is that it needs to deploy to a *different* repo. Is that the case here or are we just deploying to the gh-pages branch?

melissawm:
@jni is right, this can be simplified. Unfortunately I won't have time this week to work on this PR but I can take a look next week if it's still pending. Thanks for the fixes @GenevieveBuckley !

GenevieveBuckley:
I've got it working now - you can see a working test website here: https://genevievebuckley.github.io/napari-animation-copy/gallery.html

In the end it was simplest to add a command in the Makefile to copy the video files across into docs/_build/gallery/images, then point the actions/upload-pages-artifact to just the docs/_build folder (which now contains everything we need).

.github/workflows/deploy_docs.yml
```diff
@@ -0,0 +1,60 @@
...
+concurrency:
+  group: deploy_docs
+  cancel-in-progress: true
+
+jobs:
+  build-and-deploy:
+    name: Build & Deploy
+    runs-on: ubuntu-latest
+
+    permissions:
```
GenevieveBuckley, line 22: I'm not entirely sure if we need all three of these write permissions. I was mix-and-matching from a few different examples, so it's possible they're not all needed now.

GenevieveBuckley:
It works! 🎉 🚀 
https://napari.org/napari-animation/gallery.html

.github/workflows/deploy_docs.yml
GenevieveBuckley: Update: yes, we absolutely need all of these permissions here.

psobolewskiPhD:
Awesome! 🎉

PR #209, opened 2024-04-16
add lxml_html_clean to testing in setup.cfg

psobolewskiPhD:
Tests are failing because pytest appears to import a lot? so it picks up missing lxml_html_clean from napari
https://github.com/napari/napari-animation/actions/runs/8697301919/job/23852283978#step:7:108

Adding it to `testing` because it's not needed by anything in napari-animation. Actual use in napari is gated by try:
https://github.com/napari/napari/blob/dec8b09184728b632945ba09c7e6c8fe6c953508/napari/utils/notebook_display.py#L6-L9
But it's needed for napari tests so it's in napari[testing] but the fix isn't released yet.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -44,6 +44,7 @@ testing =
     pytest
     pytest-cov
     pytest-qt
+    # should only be needed till napari 0.5.0 release
     lxml_html_clean
 doc =
     sphinx>6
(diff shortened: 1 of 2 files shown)


psobolewskiPhD:
I think it's needed in tox.ini too?

DragaDoncila (approved):
It's alliiiiiive

Issue #213, opened 2024-05-24
Fix failing CI tests


GenevieveBuckley:
The CI tests are currently failing.

### Problem description
See here for an example: https://github.com/napari/napari-animation/actions/runs/9218650940

The reason for this failure is that `imageio-ffmpeg` does not include the FFMPEG binary for Apple Silicon Macs (see issue here https://github.com/imageio/imageio-ffmpeg/issues/71). Since the github actions runners recently switched over so that `macos-latest` runs on Apple Silicon. That means we now get an error that `imageio-ffmpeg` cannot find the FFMPEG executable.

### Proposed solution
Update the github actions workflow to check if it is running on Mac, and if it is then brew install ffmpeg (with [this github action](https://github.com/marketplace/actions/github-action-for-homebrew). We can link to https://github.com/imageio/imageio-ffmpeg/issues/71, and note this is a temporary workaround that can be removed when `imageio-ffmpeg` releases a fix.

PR #214, opened 2024-05-24
Install FFMPEG with brew on Mac CI runners

GenevieveBuckley:
Closes https://github.com/napari/napari-animation/issues/213

This is a temporary fix for issue https://github.com/imageio/imageio-ffmpeg/issues/71, which has been causing our CI to fail.


.github/workflows/test_and_deploy.yml
```diff
@@ -32,12 +32,26 @@ jobs:
...
         with:
           python-version: 3.8
+
+        # Temporary fix for 'pip install imageio-ffmpeg'
+        # not including the FFMPEG binary on Apple Silicon macs
+        # This step can be removed when issue is fixed in imageio-ffmpeg
+        # https://github.com/imageio/imageio-ffmpeg/issues/71
+      - name: brew install ffmpeg
+        if: runner.os == 'macOS'
+        uses: artemnovichkov/action-homebrew@0.1.0
```
Czaki, line 42: Are you sure that this action is needed? 
As I remember, the brew is preinstalled. In this action, brew is used without installation https://github.com/4DNucleome/PartSegCore-compiled-backend/blob/master/.github/workflows/wheels.yml

GenevieveBuckley:
The ruff CI check failed earlier because I updated ruff's target version of python from 3.8 to 3.12. To keep things simple, I've reverted that change here for now, and made a separate PR to update ruff https://github.com/napari/napari-animation/pull/218.

psobolewskiPhD:
Thanks for taking a look at this!

The pyside changes (at a minimum) from here:
[`23cdaa8` (#214)](https://github.com/napari/napari-animation/pull/214/commits/23cdaa8f74c6bd105a3f8e8ef863c1679bd8b13b)
will need to be reverted:
The latest pyside2 on pypi is 5.15.2.1
https://pypi.org/project/PySide2/
Which doesn't (officially) support even python 3.10, let alone 3.11 and 3.12:
https://wiki.qt.io/Qt_for_Python
image

Also, to echo Grzegorz's question, why not use the homebrew installed on the macOS runners?
The action you are using 
https://github.com/AnimMouse/setup-ffmpeg
Does not support arm64 as far as I can tell:
https://evermeet.cx/ffmpeg/apple-silicon-arm

The other option/alternative PR is to switch to macOS-13 instead of latest (the x86 runners).
And then maybe we consider a different imageio plugin with better wheel support: 
https://github.com/napari/napari-animation/issues/217

GenevieveBuckley:
> why not use the homebrew installed on the macOS runners?

I saw several brew actions available on the github marketplace, and assumed that meant the default runners did not come with it already installed.

GenevieveBuckley:
I'm going to close this - even though I've removed the marketplace github action (and this is change can be seen in the workflow file), the action is still being run on the CI jobs. It's probably easiest to delete this, instead of figuring out if there is a cache somewhere.

GenevieveBuckley:
Superseded by https://github.com/napari/napari-animation/pull/222

PR #222, opened 2024-06-04
Fix failing CI tests (ffmpeg on apple silicon, PySide2 problems)

GenevieveBuckley:
In this PR:
1. `brew install ffmpeg` for Apple Silicon Mac arm64 runners on github actions, since imageio-ffmpeg does not provide ffmpeg binaries for Apple Silicon https://github.com/imageio/imageio-ffmpeg/issues/71
2. Restrict PySide tests to Mac Intel (not Apple Silicon), and for all operating systems python>=3.10. PySide is not available for these configurations.
3. Drop python 3.8, and add python 3.12 to the CI test matrix

### DIFF
--- a/.github/workflows/test_and_deploy.yml
+++ b/.github/workflows/test_and_deploy.yml
@@ -42,6 +42,7 @@ jobs:
     name: ${{ matrix.platform }} py${{ matrix.python-version }}
     runs-on: ${{ matrix.platform }}
     strategy:
+      fail-fast: false
       matrix:
         platform: [ ubuntu-latest, windows-latest, macos-latest ]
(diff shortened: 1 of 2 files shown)


.github/workflows/test_and_deploy.yml
```diff
@@ -42,9 +42,10 @@ jobs:
     name: ${{ matrix.platform }} py${{ matrix.python-version }}
     runs-on: ${{ matrix.platform }}
     strategy:
+      fail-fast: false
```
GenevieveBuckley, line 45:
```suggestion
      fail-fast: true
```

jni:
Nice, thanks @GenevieveBuckley!

Issue #231, opened 2024-09-05
Replacing keyframes breaks keyframe list


psobolewskiPhD:
Tests are failing on CI:
https://github.com/napari/napari-animation/actions/runs/10724068457/job/29738857108?pr=230#step:8:123

Can reproduce locally with:
```
pytest napari_animation/_qt/_tests/test_animation_widget.py
```

Bisect points to this napari commit: 3c7d1a463df65e0e3b9668d4647cb96d066cda73
https://github.com/napari/napari/pull/7150

Edit:
So this isn't just a test fail. I can reproduce this locally using the plugin.
If you attempt to replace a keyframe (alt-r) it will:
- remove the keyframe
- add a new keyframe with the counter incremented
- Give a traceback (same as CI): ValueError: Cannot select item that is not in list: <KeyFrame: Key Frame 0>

<details>

```
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
File ~/Documents/dev/napari/napari/_qt/containers/_base_item_view.py:78, in _BaseEventedItemView.selectionChanged(self=<napari_animation._qt.keyframeslist_widget.KeyFramesListWidget object>, selected=<PyQt5.QtCore.QItemSelection object>, deselected=<PyQt5.QtCore.QItemSelection object>)
     75 desel = {i.data(ItemRole) for i in deselected.indexes()}
     77 if not self._root.selection.events.changed._emitting:
---> 78     self._root.selection.update(sel)
        sel = {<KeyFrame: Key Frame 1>}
        self._root = [<KeyFrame: Key Frame 1>]
        self = <napari_animation._qt.keyframeslist_widget.KeyFramesListWidget object at 0x32ff5d750>
     79     self._root.selection.difference_update(desel)
     80 return super().selectionChanged(selected, deselected)

File ~/Documents/dev/napari/napari/utils/events/containers/_set.py:118, in EventedSet.update(self=Selection(set()), others={<KeyFrame: Key Frame 1>})
    116 to_add = {self._pre_add_hook(i) for i in to_add}
    117 self._set.update(to_add)
--> 118 self._emit_change(added=set(to_add), removed=set())
        to_add = {<KeyFrame: Key Frame 1>}
        self = Selection(set())

File ~/Documents/dev/napari/napari/utils/events/containers/_selection.py:80, in Selection._emit_change(self=Selection(set()), added={<KeyFrame: Key Frame 1>}, removed=set())
     78 if removed is None:
     79     removed = set()
[...]

PR #233, opened 2024-09-05
Drop inserting in capture (changing in-place) and instead just add and then remove frames

psobolewskiPhD:
Closes: https://github.com/napari/napari-animation/issues/231

So in https://github.com/napari/napari/pull/7150 event handling order was changed in napari.
This is making the key frame list manipulations for inserting a replacement frame not work correctly when the list change event callbacks trigger. I tried passing `position='first'` in the `.connect()` in various places and could not get it to resolve properly -- it's events of the  SelectableEventedList, which are on the napari side I think.

I also tried to keep the name of the frame when `insert=False`, but the hash is still different so selection still fails.

In this PR I work around all this event stuff by removing the `insert` kwarg on `capture_frame` and no longer edit the frame in-place when replacing. Instead, the new frame is added and previous is removed.
Note that the frame number was already incrementing even when replacing a frame (insert=False) and because one can re-order frames, the numbers lose meaning.

### DIFF
--- a/napari_animation/_qt/animation_widget.py
+++ b/napari_animation/_qt/animation_widget.py
@@ -114,7 +114,13 @@ class AnimationWidget(QWidget):
 
     def _replace_keyframe_callback(self, event=None):
         """Replace current key-frame with new view"""
-        self.animation.capture_keyframe(**self._input_state(), insert=False)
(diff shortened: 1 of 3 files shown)


alisterburt:
Superseded by #234 - thanks for delving though!

PR #239, opened 2025-07-14
Add Python 3.13 support and fix enum behavior difference

### LINKED ISSUE #237: napari 0.6.2, py313, pyqt6: Easing related traceback, plugin can't be used
Hit this at SciPy!
```
---------------------------------------------------------------------------
KeyError                                  Traceback (most recent call last)
File ~/micromamba/envs/napari-062/lib/python3.13/site-packages/napari_animation/_qt/animation_widget.py:113, in AnimationWidget._capture_keyframe_callback(self=<napari_animation._qt.animation_widget.AnimationWidget object>, event=False)
    111 def _capture_keyframe_callback(self, event=None):
    112     """Record current key-frame"""
--> 113     self.animation.capture_keyframe(**self._input_state())
        self.animation = <napari_animation.animation.Animation object at 0x16a6d34d0>
        self = <napari_animation._qt.animation_widget.AnimationWidget object at 0x16e1419f0>
File ~/micromamba/envs/napari-062/lib/python3.13/site-packages/napari_animation/_qt/animation_widget.py:108, in AnimationWidget._input_state(self=<napari_animation._qt.animation_widget.AnimationWidget object>)
    104 def _input_state(self):
    105     """Get current state of input widgets as {key->value} parameters."""
    106     return {
    107         "steps": int(self.frameWidget.stepsSpinBox.value()),
--> 108         "ease": self.frameWidget.get_easing_func(),
        self.frameWidget = <napari_animation._qt.frame_widget.FrameWidget object at 0x12744dbd0>
        self.frameWidget.stepsSpinBox = <PyQt6.QtWidgets.QSpinBox object at 0x3235dafd0>
        self = <napari_animation._qt.animation_widget.AnimationWidget object at 0x16e1419f0>    109     }

File ~/micromamba/envs/napari-062/lib/python3.13/site-packages/napari_animation/_qt/frame_widget.py:69, in FrameWidget.get_easing_func(self=<napari_animation._qt.frame_widget.FrameWidget object>)
     68 def get_easing_func(self):
---> 69     return Easing[self.easeComboBox.currentText().upper()]
        self.easeComboBox = <PyQt6.QtWidgets.QComboBox object at 0x3235daad0>
        self = <napari_animation._qt.frame_widget.FrameWidget object at 0x12744dbd0>
File ~/micromamba/envs/napari-062/lib/python3.13/enum.py:793, in EnumType.__getitem__(cls=<enum 'Easing'>, name='')
    789 def __getitem__(cls, name):
    790     """
    791     Return the member matching `name`.
    792     """
--> 793     return cls._member_map_[name]
        cls._member_map_ = {}
        name = ''
        cls = <enum 'Easing'>
KeyError: ''
```

The Easing combobox is empty.

willingc:
Closes #237.  This PR adds Python 3.13 support.

Co-authored with @psobolewskiPhD using #240 code changes for CI.

## Key Changes

  1. Enum Compatibility Layer (`napari_animation/_enum_compat.py`):
    - Python 3.13 introduced stricter enum handling requiring `enum.member()` wrapper for callable values
    - Created compatibility layer that detects Python version and conditionally wraps enum values
    - Updated Easing and Interpolation enums to use this wrapper
  2. CI/CD Updates:
    - Added Python 3.13 to test matrix
    - Updated GitHub Actions from v4 to v5
  3. Development Tool Updates:
    - Black updated to v25.1.0 with py313 target
    - Added skip-string-normalization to preserve quote styles
    - Fixed ruff linting and Black formatting issues
  4. Test Infrastructure:
    - Added comprehensive enum compatibility tests
    - Updated mocking for Python 3.10 compatibility
  5. Drop support for 3.9

  The changes ensure napari-animation works seamlessly across Python 3.10-3.13 while maintaining backward compatibility.

### DIFF
--- a/napari_animation/_enum_compat.py
+++ b/napari_animation/_enum_compat.py
@@ -17,19 +17,20 @@ try:
     from enum import member
 
     # member is importable (Python 3.11+)
-    _HAS_ENUM_MEMBER = True
+    _ENUM_MEMBER_AVAILABLE = True
 except ImportError:
     # Python 3.10 has no member support
-    _HAS_ENUM_MEMBER = False
+    _ENUM_MEMBER_AVAILABLE = False
 
 # Check if enum.member is required (Python 3.13+)
-_NEEDS_ENUM_MEMBER = sys.version_info >= (3, 13)
+_ENUM_MEMBER_REQUIRED = sys.version_info >= (3, 13)
 
 
 def wrap_enum_member(value):
     """Conditionally wrap a value with enum.member if needed."""
-    if _NEEDS_ENUM_MEMBER:
-        # Needed for Python 3.11+
+    # Enum member is required for Python 3.13+, and available for 3.11+
+    if _ENUM_MEMBER_REQUIRED or _ENUM_MEMBER_AVAILABLE:
         return member(value)
-    # python 3.10 common case
-    return value
+    # For python 3.10 and older, just return the value directly
+    if not _ENUM_MEMBER_AVAILABLE:
+        return value
--- a/napari_animation/_enum_compat.py
+++ b/napari_animation/_enum_compat.py
@@ -29,9 +29,6 @@ _NEEDS_ENUM_MEMBER = sys.version_info >= (3, 13)
 def wrap_enum_member(value):
     """Conditionally wrap a value with enum.member if needed."""
     if _NEEDS_ENUM_MEMBER:
-        if not _HAS_ENUM_MEMBER:
-            # python 3.10 does not have member so we cannot wrap
-            return value
         # Needed for Python 3.11+
         return member(value)
     # python 3.10 common case
--- a/napari_animation/_enum_compat.py
+++ b/napari_animation/_enum_compat.py
@@ -1,13 +1,24 @@
-"""Compatibility layer for enum behavior across Python versions."""
+"""Compatibility layer for enum behavior across Python versions.
+
+Python 3.13 introduced stricter enum handling requiring `enum.member()` wrapper for callable values.
+This compatibility layer detects Python version and conditionally wraps enum values.
+- Python 3.10:
+- Python 3.11 added `member`
+- Python 3.13+: No wrapper required.
+
+Enum members have a name and a value.
+"""
 
 import sys
 
-# Check if enum.member is available (Python 3.11+)
+# Check if member is importable
 try:
     from enum import member
 
+    # member is importable (Python 3.11+)
     _HAS_ENUM_MEMBER = True
 except ImportError:
+    # Python 3.10 has no member support
     _HAS_ENUM_MEMBER = False
 
 # Check if enum.member is required (Python 3.13+)
@@ -16,6 +27,11 @@ _NEEDS_ENUM_MEMBER = sys.version_info >= (3, 13)
 
 def wrap_enum_member(value):
     """Conditionally wrap a value with enum.member if needed."""
-    if _NEEDS_ENUM_MEMBER and _HAS_ENUM_MEMBER:
+    if _NEEDS_ENUM_MEMBER:
+        if not _HAS_ENUM_MEMBER:
+            raise RuntimeError(
+                f"Python {sys.version_info[:2]} requires enum.member but it's not available. "
+                "Please upgrade your Python installation."
+            )
         return member(value)
     return value
--- a/napari_animation/_enum_compat.py
+++ b/napari_animation/_enum_compat.py
@@ -34,3 +34,5 @@ def wrap_enum_member(value):
             return value
         # Needed for Python 3.11+
         return member(value)
+    # python 3.10 common case
+    return value
--- a/napari_animation/_enum_compat.py
+++ b/napari_animation/_enum_compat.py
@@ -4,7 +4,7 @@ Python 3.13 introduced stricter enum handling requiring `enum.member()` wrapper 
 This compatibility layer detects Python version and conditionally wraps enum values.
 - Python 3.10:
 - Python 3.11 added `member`
-- Python 3.13+: No wrapper required.
+- Python 3.13+: member is required.
 
 Enum members have a name and a value.
 """
--- a/napari_animation/_enum_compat.py
+++ b/napari_animation/_enum_compat.py
@@ -1,4 +1,5 @@
 """Compatibility layer for enum behavior across Python versions."""
+
 import sys
 
 # Check if enum.member is available (Python 3.11+)
--- a/napari_animation/_enum_compat.py
+++ b/napari_animation/_enum_compat.py
@@ -1,10 +1,11 @@
 """Compatibility layer for enum behavior across Python versions.
 
-Python 3.13 introduced stricter enum handling requiring `enum.member()` wrapper for callable values.
+Python 3.13 introduced stricter enum handling requiring `enum.member()` wrapper for callable values. See https://docs.python.org/3/library/enum.html.
 This compatibility layer detects Python version and conditionally wraps enum values.
 - Python 3.10:
 - Python 3.11 added `member`
-- Python 3.13+: member is required.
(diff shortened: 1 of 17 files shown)


napari_animation/_enum_compat.py
```diff
@@ -0,0 +1,37 @@
+"""Compatibility layer for enum behavior across Python versions.
+
+Python 3.13 introduced stricter enum handling requiring `enum.member()` wrapper for callable values.
+This compatibility layer detects Python version and conditionally wraps enum values.
+- Python 3.10:
+- Python 3.11 added `member`
+- Python 3.13+: No wrapper required.
```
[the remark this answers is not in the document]
psobolewskiPhD: Yes, that is my understanding as well. py313 requires using enum.member, but py310 doesn't have that hence the wrapper being needed.
Once we drop 3.10, then we won't need the compatibility layer and can just use `member(partial(...))` from `enum`

napari_animation/_tests/test_enum_compat.py
```diff
@@ -0,0 +1,190 @@
...
+
+        def __call__(self, *args):
+            return self.value(*args)
+
+    result = TestEnum.TEST(5)
+    assert result == 10
+
+
+@pytest.mark.skip(reason="Troubleshooting")
+def test_wrap_enum_member_python_310_simulation():
```
psobolewskiPhD, line 56: I don't quite get this test. We run on py310 and py313, so the other tests should cover both behaviors?

napari_animation/_tests/test_enum_compat.py
```diff
@@ -0,0 +1,190 @@
...
+    ):
+        wrapped = wrap_enum_member(partial_func)
+
+        # Should return the original partial function
+        assert wrapped is partial_func
+        assert wrapped(5) == 10
+
+
+@pytest.mark.skip(reason="Not implemented yet")
+def test_wrap_enum_member_python_313_simulation():
```
psobolewskiPhD, line 77: Same as above, I don't get these simulation tests.

tox.ini
```diff
@@ -1,25 +1,25 @@
...
+    macos_x86_64: darwin-x86_64
+    linux: linux
+    windows: win32
 
 [testenv]
 platform = 
-    macos: darwin
-    macosintel: darwin
+    macos_arm64: darwin-arm64
+    macos_x86_64: darwin-x86_64
```
psobolewskiPhD, line 22: This should be revered, the platform is just `darwin` as far as I know

tox.ini
```diff
@@ -1,25 +1,25 @@
...
 [gh-actions:env]
 PLATFORM =
-    ubuntu-latest: linux
-    macos-latest: macos
-    macos-13: macosintel
-    windows-latest: windows
+    macos_arm64: darwin-arm64
+    macos_x86_64: darwin-x86_64
+    linux: linux
+    windows: win32
```
psobolewskiPhD, line 17: I think this needs to be reverted, because this is supposed to map the GitHub runner (`platform`) to the tox env.
The runners are defined in test_and_deploy:
`platform: [ ubuntu-latest, windows-latest, macos-latest ]`
So looks like we dropped testing on macosintel and could just remove that.


Or it needs to be changed to have the GitHub names on the left hand side...

tox.ini
```diff
@@ -36,20 +36,26 @@ deps =
...
 deps =
     {[testenv]deps}
     napari[pyside2]
-    lxml_html_clean # should only be needed till napari 0.5.0
-    # .  # napari-animation install from source
+
+[testenv:py{310}-{linux,macos_x86_64,macos_arm64,windows}-pyside6]
+deps =
+    {[testenv]deps}
+    napari[PySide6]
```
psobolewskiPhD, line 58: this doesn't exist (yet)

tox.ini
```diff
@@ -36,20 +36,26 @@ deps =
...
-    lxml_html_clean # should only be needed till napari 0.5.0
-    # .  # napari-animation install from source
+    napari[PyQt6]
 
 # PySide2 unavailable on Mac apple silicon arm64, and for python >=3.11
-[testenv:py{39,310}-{linux,macosintel,windows}-pyside]
+[testenv:py{310}-{linux,macos_x86_64,windows}-pyside2]
 deps =
     {[testenv]deps}
     napari[pyside2]
```
psobolewskiPhD, line 53: pyside2 requires `pytest-qt < 4.5.0`

psobolewskiPhD (dismissed):
I think the enum compatibility is a very clever fix till we drop 3.10! Thanks!
but there are some CI issues with the tox.ini changes.
I've tried to break out just fixing CI and running on py313 into their own PR:
https://github.com/napari/napari-animation/pull/240
Hopefully that makes this PR easier!

napari_animation/_enum_compat.py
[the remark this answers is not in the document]
jni: Should we just drop 3.10 then? [SPEC0 says yes](https://scientific-python.org/specs/spec-0000/#support-window).

napari_animation/_enum_compat.py
```diff
@@ -0,0 +1,37 @@
...
+
+
+def wrap_enum_member(value):
+    """Conditionally wrap a value with enum.member if needed."""
+    if _NEEDS_ENUM_MEMBER:
+        if not _HAS_ENUM_MEMBER:
+            raise RuntimeError(
+                f"Python {sys.version_info[:2]} requires enum.member but it's not available. "
+                "Please upgrade your Python installation."
+            )
```
jni, line 35: This code should be unreachable, so I am in favour of removing it (and the corresponding test)

napari_animation/_enum_compat.py
```diff
@@ -0,0 +1,37 @@
+"""Compatibility layer for enum behavior across Python versions.
+
+Python 3.13 introduced stricter enum handling requiring `enum.member()` wrapper for callable values.
```
jni, line 3: Can we link the relevant PEPs or Python docs here?

napari_animation/_enum_compat.py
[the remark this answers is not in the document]
jni: A fair point! I'm happy to leave it. You're making me want to follow SPEC0 more strictly on napari main, though 😜. 

I think though that we should be even clearer about the bits of code that can be removed. (See this discussion for example.) I think the whole `_enum_compat` file should be able to go once we drop 3.10? Which will be in a few months with napari 0.7.0.

willingc:
@jni @psobolewskiPhD @brisvag Maybe we should simplify everything and drop 3.10 early. We can then drop all of the enum guarding which is a PITA to maintain. I would be inclined to also move this to a modern packaging deployment too. Thoughts?

jni:
I actually think we don't need to worry *too much* about simul-dropping — if you are on 3.10 you'll simply get a slightly older napari-animation, but it'll work. Right?

willingc:
@jni I added back 3.10 for pyside for linux and window (no mac since I couldn't get it to work). I also updated the ci to output in the run name to have the qt_version displayed.

Let me go back (now that CI is green) and check the nested guard that you mention. Honestly, I spent most of my time fighting with tox and CI, as they were not as explicit as I would have expected. Thankfully, @psobolewskiPhD had the fail-fast toggled to false in his PR which ultimately unblocked me.

willingc:
@jni Removed the nested check and improved the naming of the guard variables. This should be ready to land. Thanks for the feedback.

psobolewskiPhD:
Just a quick note: we really shouldn't drop py310 in this repo, because it will make it incompatible with pyside2. For the time being, we need to keep that for the bundle, until we move everything over to pyside6.
This is a key plugin that needs to work in the bundle.

jni:
@psobolewskiPhD the old version of napari-animation would still be available for the bundle *if* the bundle was on 3.10! But it's actually on 3.11!

https://github.com/napari/packaging/pull/189

Issue #225, opened 2020-05-25
Add screen recording functions for napari


GenevieveBuckley:
## 🚀 Feature

It would be nice if we had screen recording functionality for the napari viewer (this was first suggested in https://github.com/napari/napari/issues/1267).

This feature is dependent on https://github.com/napari/napari/pull/780, so that needs to be merged before any work starts here.

From Juan's comment ([link](https://github.com/napari/napari/issues/1267)):
> if you look at the the IPython display documentation:
> https://ipython.readthedocs.io/en/stable/api/generated/IPython.display.html
> there's also the ability to display a video. So it would even be possible to do e.g. `napari.record(True)`, then `nbrecording(v)` would embed a video of the viewer's interactions.

Other context:
We might like to match the structure of `screenshot()` & `nbscreenshot()`, see https://github.com/napari/napari/pull/1269

psobolewskiPhD:
Something to keep an eye on from the prolific @kephale 
https://github.com/kephale/napari-screen-recorder

kephale:
I could contribute that kind of functionality to napari-animation, but I really just wanted something simple to make things like this https://mastodon.social/@kisharrington/112905891764696439

psobolewskiPhD:
💯 @kephale 
I was just reminded of this thread and thought your new plugin could be useful for someone who comes across it.

PR #260, opened 2026-04-16
move to src layout and modernize to most recent copier template

brisvag:
Following up on #225. I re-created the whole repo based on the most recent napari plugin template, and then moved everything where it needed to be. Hopefully nothing broke :P


willingc:
Oop just saw this one. Can we liberate `_tests` to `tests` at the top-level? This seems a better pattern for plugins.

And huge +1 for modernizing!

willingc (approved):
Reran the failed test. This looks good.

@brisvag Do you want to merge it now?

I still would love to see us move the tests directory to the root level but that could definitely be a different PR.

TimMonko:
> Oop just saw this one. Can we liberate `_tests` to `tests` at the top-level? This seems a better pattern for plugins.
> 
> And huge +1 for modernizing!

This is also the current layout of the napari-plugin-template ;)

TimMonko (approved):

brisvag:
Let's merge to reduce diffs, I'll follow up with the test move!

PR #261, opened 2026-04-16
Timeline widget

brisvag:
# Description

This adds a widget with tracks and keyframes to give higher control to the animation machinery.

https://github.com/user-attachments/assets/80d9a278-8aa1-4bbe-8113-9504de785635

Unfortunately the current code here is very hard-wired around keyframes being full viewer states, so this new animation widget is effectively fully separate.

In my opinion, if we get the api cleaned up, this new widget and backend can be a full replacement of the current code, giving more control over the general structure of the animation, as well as more interactivity to the widget (e.g: already has #100 plus other quality of life things like looping/pingpong and selecting a replay range).

The current behaviour (single keyframe for the whole viewer), which is very simple and intuitive for simple animations, can still be replicated by selecting the `viewer` track, which interpolates the whole viewer, with the added benefit of being able to interactively move the keyframes and the stuff mentioned above!

---

Instead of making a a plugin (https://github.com/brisvag/napari-animation-timeline), opening a PR here to see if we can do a proper integration.

Note that this PR includes #260, so the diff looks crazy, but basically I only added 1 file (`_qt/timeline_widget.py`).

The bulk of the code is napari-agnostic (almost...) and lives here: github.com/brisvag/qt-animation-timeline/


willingc:
Yikes, this is a big PR. If you are able, it might make sense to split out general cleanup and edits from the widget.

brisvag:
Merged #260, now it's all nice and clean!

src/napari_animation/_qt/timeline_widget.py
```diff
@@ -0,0 +1,250 @@
...
+            _LAYER_TRACK_OPTIONS,
+            strict=True,
+        ):
+            new_name = name_template.format(layer_name=layer.name)
+            self.timeline.animation.rename_track(old_name, new_name)
+            new_opts[new_name] = old_val
+
+        self.layer_track_options[layer] = new_opts
+
+    def add_custom_track(self, name: str, model: Any, attr: str):
```
willingc, line 151: Would it make sense to add a test for a custom track?

src/napari_animation/_qt/timeline_widget.py
```diff
@@ -0,0 +1,250 @@
...
+from qt_animation_timeline import (
+    AnimationTimelineWidget as _AnimationTimelineWidget,
+)
+from qt_animation_timeline import PlayMode
+from qtpy.QtWidgets import QGridLayout, QLabel, QPushButton, QSpinBox, QWidget
+
+if TYPE_CHECKING:
+    import napari
+
+# centralized whitelisted set of track options.
```
willingc, line 26: I'm not quite sure from the comment if "track options" only applies to the timeline widget or beyond to the whole viewer.

Similarly, for _LAYER_TRACK_OPTIONS

willingc (commented):
A few small things...

@brisvag Looks awesome.

src/napari_animation/_qt/timeline_widget.py
brisvag: I'm not sure I get the question :thinking: What do you mean to "apply to the whole viewer"? This is the pre-generated set of tracks that will be available in the dropdown by default for the viewer and for each layer.

psobolewskiPhD:
One concern: your package has a dependency on pyqt6, which would make napari-animation depend on pyqt6, which would break e.g. the bundle. I think napari-animation is very much a flagship plugin -- which doesn't get enough press! -- so this concerns me. Is there a way you can make things work with pyside6?

src/napari_animation/_qt/timeline_widget.py
brisvag: Done!

brisvag:
Oh, that was just a mistake :thinking: I thought I did everything with qtpy....

Yep indeed everything works. I just put it in the runtime deps by mistake instead of the test ones!

brisvag:
Updated the qt package on pypi to remove the unnecessary pyqt6 dep and with the links by @psobolewskiPhD!

willingc:
We missed moving the `_qt` tests out of src in #264 @brisvag

brisvag:
Move those as well, and now this can both read and write movies and export/import the animation itself as json!

src/napari_animation/_qt/timeline_widget.py
```diff
@@ -0,0 +1,313 @@
...
+            frames per second
+        quality: float
+            number from 1 (lowest quality) to 9
+            only applies to non-gif extensions
+        file_format: str
+            The format to use to write the file. By default imageio selects the appropriate
+            for you based on the filename.
+        canvas_only : bool
+            If True include just includes the canvas, otherwise include the full napari
+            viewer.
```
willingc, line 224:
```suggestion
            If True, save only includes the canvas; otherwise, save includes the full napari
            viewer.
```

src/napari_animation/_qt/timeline_widget.py
```diff
@@ -0,0 +1,313 @@
...
+        frames = []
+
+        for _ in cancelable_progress(
+            frame_iterator, desc='Rendering animation...', total=anim.n_frames
+        ):
+            image = self.viewer.screenshot(
+                canvas_only=canvas_only, scale=scale_factor, flash=False
+            )
+            frames.append(image)
+        if mode == PlayMode.PINGPONG:
```
willingc, line 292:
```suggestion

        if mode == PlayMode.PINGPONG:
```

src/napari_animation/_qt/timeline_widget.py
```diff
@@ -0,0 +1,313 @@
...
+            frames = frames + frames[::-1]
+
+        for frame in frames:
+            writer.append_data(frame)
+
+        anim.play_mode = mode
+
+    def save_timeline(self, filename):
```
willingc, line 300:
```suggestion
    def save_timeline(self, filename):
        """Save timeline to a file given a filename"""
```

src/napari_animation/_qt/timeline_widget.py
```diff
@@ -0,0 +1,313 @@
...
+        dump = self.timeline.animation.model_dump_json(indent=4)
+        path = Path(filename)
+        with open(path, 'w') as f:
+            f.write(dump)
+
+    def load_timeline(self, filename):
```
willingc, line 306:
```suggestion
    def load_timeline(self, filename):
        """Load a timeline from a file given a filename"""
```

src/napari_animation/_qt/timeline_widget.py
```diff
@@ -0,0 +1,313 @@
...
+    def save_movie(
+        self,
+        filename,
+        fps=20,
+        quality=5,
+        file_format=None,
+        canvas_only=True,
+        scale_factor=None,
+    ):
+        """Create a movie based on key-frames
```
willingc, line 206:
```suggestion
        """Create and save a movie based on key-frames
```

willingc (approved):
Looking good @brisvag. A few suggestions on adding small docstrings for IDE tooltips.

willingc:
Let's try pinning Pydantic in tox.ini for 3.10 and pyside2


`pydantic<2.11  # psygnal used by napari expects older pydantic internals; avoid breakage on newer pydantic`

brisvag:
Mmm I think the problem is that `qt-animation-timeline` uses `pydantic>2` stuff, but napari <0.7.0 does not support pydantic v2 properly, so we get conflicts... Though I'm confused about why the errors are so unclear.

brisvag:
I don't undersdtand this failure, no matter what I try :/

brisvag:
Yay, finally got tests running an passing! Thanks @Czaki. Now this should be good for a final review/merge!

